PR #1, opened 2020-12-03
Initialize repo and add data download script

jjc2718:
These changes initialize the repo for our multi-modal mutation prediction project, and add a script to download various TCGA data modalities from GDC.

### DIFF
--- a/00_download_data/nbconverted/download_data.py
+++ b/00_download_data/nbconverted/download_data.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Download multiple modalities of pan-cancer data from TCGA
+# 
+# The data is accessed directly from the [Genome Data Commons](https://gdc.cancer.gov/about-data/publications/pancanatlas).
+
+# In[1]:
+
+
+import os
+import pandas as pd
+from urllib.request import urlretrieve
+
+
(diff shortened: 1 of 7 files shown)


nrosed:
Hi Jake. It looks fine for me, I only have some concerns on commit d196e0d. I am not a python nor notebook expert so I am not sure how the file `/nbconverted/download_data.py` is intended to be used in the future. However, if it is supposed to be run by someone in lieu of the notebook, I would suggest to wrap the code into a `main` function, and separate all the data reading sections into methods.

00_download_data/nbconverted/download_data.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
```
nrosed, line 1: As mentioned in my comment, I would suggest to wrap the code into a main function, and separate all the data reading sections into methods.

jjc2718: Thanks @nrosed! I (and others in the lab) generally export notebooks to Python files to make code review a bit easier, since it's hard to read a diff of the raw JSON in the `.ipynb` files. Since this is a new notebook it doesn't matter as much, but for changes to existing notebooks it can be hard to tell what actually changed and where to make inline comments.

This is a good point, though - someone new to the repo could easily be confused by this. I'll add a note to the README making it clear that this download script is intended to be run as a notebook in a linear fashion.

nrosed: aha, ok makes sense. Thanks for the info!

PR #2, opened 2020-12-09
Gene expression/methylation data preprocessing and exploration

jjc2718:
The main goal of this PR is to preprocess the methylation data from TCGA, and to do some basic data exploration to verify that our preprocessing isn't removing informative signal.

IDH1 is a well-known driver in a subtype of glioma (~80% of samples) that has an important role in DNA methylation, so we were fairly sure that we'd see strong clustering in the methylation data based on IDH1 mutation status (and indeed we did). This suggests that IDH1 can serve as a positive control once we start fitting models to predict mutation status based on methylation data (coming in future PRs).

**Note**: This PR isn't nearly as big as it looks. Files to focus on in review:
* `00_download_data/1B_preprocess_methylation_data.ipynb`
* `00_download_data/2_explore_data.ipynb`
* and any feedback on the general repo/file structure would be appreciated as well (as always).

The changes in `mpmp/config.py`, `mpmp/utilities/data_utilities.py`, and `mpmp/utilities/tcga_utilities.py` are copied from the `pancancer-evaluation` repo and don't need to be reviewed. The changes in `00_download_data/1A_preprocess_expression_data.ipynb` are mostly the same as the preprocessing notebook in `pancancer-evaluation` as well, so these don't need to be reviewed in detail either (unless you want to).


00_download_data/1B_preprocess_methylation_data.ipynb
```diff
@@ -0,0 +1,444 @@
...
+       "      <td>0.019061</td>\n",
+       "      <td>0.017349</td>\n",
+       "    </tr>\n",
+       "  </tbody>\n",
+       "</table>\n",
+       "</div>"
+      ],
+      "text/plain": [
+       "Composite Element REF  cg00005847  cg00008493  cg00012199  cg00012386  \\\n",
+       "sample_id                                                               \n",
```
ben-heil, line 387: I'm confused by what's going on here. Why is the sample ID on its own line? Shouldn't it just be a header column like the rest?

00_download_data/2_explore_data.ipynb
```diff
@@ -0,0 +1,854 @@
...
...JroEQQgghpK6p62MwCigRQgghdUjUlRR8vC4Gp+KTDXbMgwf3ITx8FDw9vWBhYYGEhBt49iwT7u7NlbZzdBQBwPPnPJSea9SoYRa2JIQQQkjDwI3Boq6kGOyYdX0MRgElQgghpA45EPMIOQVi7It+aJDj5efn4/z5GOzYsQUfffQ+iooKsXv3Nri6uiEj46nStv/+G4usrGdwcXFDRka60nPXrl1BcvITg7SJEEIIIaS24cZgB2IeGeR49WEMRgElQgghpA4JD/aCk4MVIkIMk9r8zz+HMWJEBFauXIcVK9Zgw4bf8e+/cRg8eBg2bvwFUqkUAJCU9BjffLMIZmZmGD48HFu2bEJJSQkAICcnG199tRClpaUGaRMhhBBCSG3DjcHCg70Mcrz6MAajotyEEEJIHRLq74FQfw8IhWaQyeTVPt6BA/swZ85C/mdra2v07RuGjIx0+Pp2wv/+9zYsLCwgl8swZ85CODk5w8nJGSNHjsaMGVNhbm4OsbgU//3vVLRt267a7SGEEEIIqY24MZih1IcxmIAxxkxy5iqSSGTIzS2u8fOIRLZGOU9dRH2jGfWNZtQ3mlHfaFbf++bp08dwdW1ZpX0NFVCqDdT1Q9OmDiZqDdHEWGMwY6nvf19qE+pr46G+Nh7qa+Opib6uzhisvjDE+IumvBFCCCGEEEIIIYQQvVBAiRBCCCGEEEIIIYTohWooEdLA5Z6OQvbBfXAeEQFR31BTN4cQQgghpMGKTonDkUfHYQYzZItz+Mcn+LyMEI8gE7aMEEIqogwlQhq47IP7IM3JQfbBfaZuCiGEEEJIg3bk0XHkivOUgkkAsPXObhO1iBBCNKOAEiENnPOICJg7OcF5RISpm0IIIYQQ0qAN9Rqg9nEGhuiUOCO3hhBCtKMpb4Q0cKK+oTTVjRBCCCGkFgjxCMK93Ae4kH65wnN77x2iaW+EkFqFAkqEEEJIAxYffxFz586El1crAIBUKsXMmXOxefNG9O8/CD169FLaPj39KdauXYWcnGyIxWL4+HTAhx9+DAsLCwBAQsINTJ36Ntav/wUdOvga/XoIIaSue9N3gtqAUomsFNEpcRRUIqSeqA9jMJryRgghhDRwgYFdsXbtBqxduwGTJ0/BunWr1G4nk8kwc+bHGD/+NaxduwE//fQ7zM3N8fPPP/DbHDy4F+PHv4bdu3cYqfWEEFL
```
ben-heil, line 820: If you plan on using the reduced-dimensionality data downstream, and you have methylation/expression data for the same samples, you might be interested in something like [joint non-negative matrix factorization](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5006236/).

ben-heil, line 829: Do you have an idea for what the two clusters of IDH1 mutant cancers might be?

mpmp/utilities/data_utilities.py
```diff
@@ -0,0 +1,218 @@
...
+            if gene in genes_df.gene:
+                classification = genes_df[genes_df.gene == gene].classification.iloc[0]
+    return classification
+
+
+def load_pancancer_data_from_repo(subset_columns=None):
+    """Load data to build feature matrices from pancancer repo. """
+
+    base_url = "https://github.com/greenelab/pancancer/raw"
+    commit = "2a0683b68017fb226f4053e63415e4356191734f"
```
ben-heil, line 180: Should these be in your config file instead of hardcoded?

ben-heil (approved):
Looks good! Bonus points if you can find a way to incorporate MCMC so your method name is mpmpmcmc

00_download_data/2_explore_data.ipynb
jjc2718: Neat idea! I'm planning to start by using the data modalities for classification independently, but I do think we'll test some ways to combine them down the road - that seems like a good way to approach it.

00_download_data/1B_preprocess_methylation_data.ipynb
jjc2718: I think that's just how the notebook formats `df.columns.name` (i.e. the columns name, or in this case it was the index name until I transposed the dataframe). The columns name doesn't get written as part of `df.to_csv`; the index name will just be a header column there like you're suggesting.

I'm not worried about this - I know the columns are methylation CpG probe IDs, and labeling them with the TCGA default `Composite Element REF` doesn't really tell me much.

00_download_data/2_explore_data.ipynb
jjc2718: Good question! I don't know a lot about glioma, honestly, but looking at Figure 1 of [this paper](https://doi.org/10.1007/s13311-017-0519-x) it seems like there are two major subclasses of IDH-mutant gliomas (oligodendrogliomas and diffuse astrocytomas).

IDH-mutant diffuse astrocytomas seem to be characterized by TP53 and ATRX mutations, so I tried looking at those, to see if they would separate nicely:

image
image

So in the methylation data there's not much of a separation, but in the gene expression data lots of samples in the upper cluster seem to have IDH1/TP53/ATRX mutations, so my guess is that those are astrocytomas and the other IDH1 mutant gene expression cluster is oligodendrogliomas. TCGA might have this subtype information (I'm not sure off the top of my head) so it could be interesting to look at that as well!

PR #8, opened 2020-12-29
Set up and run mutation classification for Vogelstein genes

jjc2718:
**Description of changes/results in PR:**

Our next step in this analysis is to compare methylation data with gene expression data for predicting mutations in cancer drivers. Like the previous analysis in the `pancancer-evaluation` repo, we're using the cancer gene set from [Vogelstein and Kinzler 2004](https://www.nature.com/articles/nm1087). We hypothesized that for most genes expression data would be the best predictor, but for some genes (particularly those that are involved in DNA methylation or that affect methylation patterns directly) the methylation data would provide better predictions.

However, we can see from this plot that this isn't really the case here:

image

This shows that expression data provides better predictions than methylation data in many cases (top right) and methylation data doesn't provide a significant improvement over expression data for any of the 95 genes we considered (top left).

**Overview of what needs to be reviewed:**

Files to focus your review on (these are mostly new):
* `01_classify_stratified/run_mutation_classification.py`
* `01_classify_stratified/plot_mutation_results.ipynb` (or converted Python script in `01_classify_stratified/nbconverted`)
* Changes to `mpmp/data_models/tcga_data_model.py`, `mpmp/utilities/analysis_utilities.py`, and `mpmp/utilities/classify_utilities.py`

Files that don't need to be looked at closely:
* `mpmp/exceptions.py`(copied from `pancancer-evaluation` repo)
* Small changes to `01_classify_stratified/run_cancer_type_classification.py` and `01_classify_stratified/plot_cancer_type_results.ipynb`


01_classify_stratified/nbconverted/plot_mutation_results.py
```diff
@@ -0,0 +1,217 @@
...
+plt.text(0.7, -np.log10(SIG_ALPHA)+0.05,
+         r'$\alpha = {}$'.format(SIG_ALPHA),
+         va='center', ha='center', color=l.get_color(),
+         backgroundcolor=plt.gca().get_facecolor())
+plt.xlabel('AUPR(expression) - AUPR(methylation)')
+plt.ylabel(r'$-\log_{10}($adjusted $p$-value$)$')
+plt.xlim((-0.75, 0.75))
+plt.ylim((0, y_max+5))
+plt.legend(title=r'Reject $H_0$', loc='upper left')
+plt.title(r'Cancer type prediction, expression vs. methylation')
```
ben-heil, line 190: It's interesting that methylation seems to have a strictly weaker signal. Do you plan on incorporating both methylation and expression into a single model to see if it outperforms the individual models?

ben-heil (approved):
Looks good! Do you think the methylation data and expression data are giving different signals? Or both saying "If you have a BRAF mutation everything is bad"?

ben-heil:
Ooh also I had never seen the "mark a PR as a draft" feature before, that's really cool

jjc2718:
> Looks good! Do you think the methylation data and expression data are giving different signals? Or both saying "If you have a BRAF mutation everything is bad"?

My guess is that it's mostly the latter (methylation and expression data are indicating similar/redundant downstream effects of a mutation, especially since methylation often controls or at least correlates with expression), but that's not really based on anything concrete. I'm definitely interested in designing some experiments to tease this apart going forward.

> Ooh also I had never seen the "mark a PR as a draft" feature before, that's really cool

Yeah! I learned it from @ajlee21 , she uses it sometimes for her works in progress. It's a neat feature!

01_classify_stratified/nbconverted/plot_mutation_results.py
jjc2718: Yeah, I'd definitely like to do this. Not sure if I'll get to this soon or later, but it's on the list of future directions for sure.

ajlee21:
And I learned it from @dongbohu and/or @cgreene, tis the circle of life!

PR #9, opened 2021-01-05
Specify model options and save to experiment directory

jjc2718:
Based on [this review comment](https://github.com/greenelab/mpmp/pull/4#discussion_r546740144), I was trying to think of ways to save model hyperparameters and other important metadata with the experiment results each time I run an experiment. The solution I came up with uses `argparse` [argument groups](https://docs.python.org/3/library/argparse.html#argument-groups) to separate model options from arguments that don't affect the output. Then I just save those as a Python dict to a pickle file, which I can reference with the results in my downstream analyses as needed.

I thought about using hdf5 files to attach the metadata directly to my results dataframes, but it seemed like more trouble than it was worth, although it looks like there are [some ways to do it](https://stackoverflow.com/q/14688306).

I definitely don't think this is a perfect solution, and I'm sure I'll iterate on it in the future, but it should be a decent start.


01_classify_stratified/run_mutation_classification.py
```diff
@@ -22,57 +22,95 @@
...
-    sample_info_df = du.load_sample_info(verbose=args.verbose)
+    io_args, model_options = process_args()
+    sample_info_df = du.load_sample_info(verbose=io_args.verbose)
 
     # create results dir if it doesn't exist
-    args.results_dir.mkdir(parents=True, exist_ok=True)
+    io_args.results_dir.mkdir(parents=True, exist_ok=True)
+    # create subdir for experiment if it doesn't exist
+    experiment_dir = Path(io_args.results_dir, 'gene').resolve()
+    experiment_dir.mkdir(parents=True, exist_ok=True)
```
miltondp, line 109: Same comment for these lines in the other file.

01_classify_stratified/run_mutation_classification.py
```diff
@@ -22,57 +22,95 @@
...
+
+    # save model options for this experiment
+    # (hyperparameters, preprocessing info, etc)
+    fu.save_model_options(experiment_dir, model_options)
```
miltondp, line 113: Just thinking aloud (minor comment): would it be helpful to have this function in `TCGADataModel` instead?

mpmp/utilities/data_utilities.py
```diff
@@ -213,3 +213,21 @@ def load_sample_info(verbose=False):
...
+            continue
+        group_dict = {
+            a.dest : getattr(args, a.dest, None) for a in group._group_actions
+        }
+        arg_groups[group.title] = argparse.Namespace(**group_dict)
+    return arg_groups
+
+
+
+
```
miltondp, line 233: Are these extra lines needed?

mpmp/utilities/file_utilities.py
```diff
@@ -2,27 +2,83 @@
...
+    signal = 'shuffled' if shuffle_labels else 'signal'
+    seed = model_options.seed
+    if identifier is None:
+        # this might be the case for files that pertain to the whole experiment
+        # (e.g. metadata/parameters file)
+        return Path(output_dir,
+                    '{}_s{}_{}{}'.format(training_data,
+                                         seed,
+                                         file_descriptor,
+                                         extension))
```
miltondp, line 39: If using Python version >= 3.6, you can use f-strings here if you like them:

```suggestion
        return Path(output_dir,
                    f'{training_data}_s{seed}_{file_descriptor}{extension}')
```

mpmp/utilities/file_utilities.py
```diff
@@ -59,23 +114,32 @@ def save_results(output_dir,
...
-        output_dir, "{}_{}_{}_s{}_auc_threshold_metrics.tsv.gz".format(
-            identifier, training_data, signal, seed)).resolve()
+    output_file = construct_filename(output_dir,
+                                     identifier,
+                                     shuffle_labels,
+                                     model_options,
+                                     'auc_threshold_metrics',
+                                     '.tsv.gz')
     auc_df.to_csv(
         output_file, sep="\t", index=False, compression="gzip", float_format="%.5g"
```
miltondp, line 124: I think pandas automatically detects the compression algorithm from the file extension provided (`.tar.gz`). If this is true, it might be better to avoid specifying `compression="gzip"` here, and just handle that using the file extension.

miltondp (approved):
Very nice solution, @jjc2718! I liked the option of having argument groups with `argparse`, I wonder if I will be able to use something like that with jupyter notebooks. I agree that `hdf5` might be more problematic.

I'm approving this PR and left some minor comments and ideas for the future.

01_classify_stratified/run_cancer_type_classification.py
```diff
@@ -22,27 +22,49 @@
...
+
+    # argument group for parameters related to input/output
+    # (e.g. filenames, logging/verbosity options, target genes)
+    #
+    # these don't affect the model output, and thus don't need to be saved
+    # with the results of the experiment
+    io = parser.add_argument_group('io',
+                                   'arguments related to script input/output, '
+                                   'note these will *not* be saved in metadata ')
+    io.add_argument('--cancer_types', nargs='*',
```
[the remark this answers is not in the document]
jjc2718: Yeah, this is kind of a subtle point - this argument affects which output is generated (each experiment predicts one cancer type), but the cancer type is in the filename when the results are generated, and other than that the combination of cancer types that the script runs for doesn't matter.

So, for example, if I run the script for breast cancer (BRCA) on its own, the results I generate for breast cancer will be the same as if I run it for BRCA, thyroid cancer (THCA) and low-grade glioma (LGG) together. In other words, each cancer type runs independently of the others. So, there's not really any need to save the list of cancer types that I ran the script for.

Does that make sense?

01_classify_stratified/run_cancer_type_classification.py
```diff
@@ -58,38 +80,55 @@ def process_args():
...
 
-    return args, sample_info_df
+    # split args into defined argument groups, since we'll use them differently
+    arg_groups = du.split_argument_groups(args, parser)
+    io_args, model_options = arg_groups['io'], arg_groups['model_options']
+
+    # add some additional hyperparameters/ranges from config file to model options
+    model_options.alphas = cfg.alphas
+    model_options.l1_ratios = cfg.l1_ratios
+    model_options.standardize_data_types = cfg.standardize_data_types
```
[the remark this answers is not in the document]
jjc2718: Correct - I'll add that to the comment.

01_classify_stratified/run_mutation_classification.py
```diff
@@ -22,57 +22,95 @@
...
+                      help='number of folds of cross-validation to run')
+    opts.add_argument('--seed', type=int, default=cfg.default_seed)
+    opts.add_argument('--subset_mad_genes', type=int, default=cfg.num_features_raw,
+                      help='if included, subset gene features to this number of '
+                           'features having highest mean absolute deviation')
+    opts.add_argument('--training_data', type=str, default='expression',
+                      choices=['expression', 'methylation'],
+                      help='what data type to train model on')
+
+    args = parser.parse_args()
```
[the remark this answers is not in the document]
jjc2718: That's a good idea! There is definitely a lot of overlap. I'll create an issue to do this in the future.

01_classify_stratified/run_cancer_type_classification.py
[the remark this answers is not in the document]
miltondp: Ah, yes, that totally makes sense!

01_classify_stratified/run_cancer_type_classification.py
```diff
@@ -58,38 +80,55 @@ def process_args():
...
         'skip_reason'
     ]
-    if args.log_file.exists() and args.log_file.is_file():
-        log_df = pd.read_csv(args.log_file, sep='\t')
+    if io_args.log_file.exists() and io_args.log_file.is_file():
+        log_df = pd.read_csv(io_args.log_file, sep='\t')
     else:
         log_df = pd.DataFrame(columns=log_columns)
-        log_df.to_csv(args.log_file, sep='\t')
+        log_df.to_csv(io_args.log_file, sep='\t')
```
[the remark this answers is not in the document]
jjc2718: Yeah! I usually write warnings/errors to a separate file, since the df isn't really flexible to handle any arbitrary type of warning or error, but the ones I expect to see I put into a dataframe.

mpmp/utilities/data_utilities.py
jjc2718: Nope, will remove

mpmp/utilities/file_utilities.py
jjc2718: I should update all of my scripts to use these - created #12 for this.

mpmp/utilities/file_utilities.py
jjc2718: Cool, seems like this works! I didn't know pandas could infer the compression algorithm automatically.

01_classify_stratified/run_mutation_classification.py
jjc2718: Yeah, that might make more sense. I'm going to leave it for now, but I'll make a note in #11 to think about this in the future (since I'll be refactoring the argument parsing/saving code anyway).

01_classify_stratified/run_cancer_type_classification.py
```diff
@@ -58,38 +80,55 @@ def process_args():
...
     # process command line arguments
-    args, sample_info_df = process_args()
+    io_args, model_options, sample_info_df = process_args()
 
     # create results dir if it doesn't exist
-    args.results_dir.mkdir(parents=True, exist_ok=True)
+    io_args.results_dir.mkdir(parents=True, exist_ok=True)
+    # create subdir for experiment if it doesn't exist
+    experiment_dir = Path(io_args.results_dir, 'cancer_type').resolve()
+    experiment_dir.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
jjc2718: > Will `io_args.results_dir` always contain a `Path` object? Not sure how `argparse` handles this. I thought you might get a `str` instance here and thus `mkdir` would fail.

The argparse input should be converted from a `str` to a `Path` object in the `process_args` function (see line 69), so I think this should work fine as is. Good to think about it though!

> Also, since you are using `parents=True`, you can safely remove line 104 I think (and avoid the potential issue I mention above).

Good catch!

PR #10, opened 2021-01-06
Plot confusion matrix for cancer type prediction

jjc2718:
**Changes in this PR:**
* Added an option to cancer type prediction script to save positive class probabilities in results directory
* Added code to `01_classify_stratified/plot_cancer_type_results.ipynb` to use this data to plot heatmaps (confusion matrices) for expression and methylation data

**Interpretation:**

The confusion matrices are pretty much what I expected based on the outcome of the prediction experiments (see https://github.com/greenelab/mpmp/pull/4). The methylation matrix is much noisier than the expression matrix, which corresponds to the decreased prediction accuracy that we observed. 

In the expression confusion matrices (see below), most of the off-diagonal points with high average probabilities make sense: COAD and READ (colon cancer/rectal cancer) are known to be extremely similar, block diagonals around the kidney cancers and uterine cancers, LGG and GBM are both brain tumors from a similar cell lineage (gliomas), esophageal and stomach cancer are similar.

image

The only thing I'm not sure about are the horizontal lines for pancreatic cancer (PAAD) and kidney chromophobe carcinoma (KICH). Maybe they're just rare cancers so the label imbalance is throwing off the model? I'll have to look further into this.


ajlee21 (approved):
Some comments on `plot_cancer_type_results.ipynb`
* Just curious, is the alpha = 0.001 standard when using human cancer data? It seems very stringent, but perhaps that is because there are many genes found so this threshold needs to be very strict.
* Very minor, do you want to add FILTER_CANCER_TYPES to your cfg file?
* So your boxplot is showing how well gene expression vs methylation data can predict if a sample is cancer vs normal? Or is this still the same task as before for predicting if a gene is mutated within some cancer type or not? OVerall it looks like methylation data doesn't perform as well compared to expression data and there is a lot more variance. Though there are some where the prediction looks to be equally high using both expression and methylation, I wonder what is happening for these cases. Looking at your scatter plots it looks like there are only a few cancer types where expression data is significantly more predictive compared to methylation data.


* Looks like you did a similar analysis comparing performance using expression vs methylation on the task of predicting mutation or not in plot_mutation_results.ipynb. Might be helpful to add a sentence at the header to describe the task here. According to your scatter plot it does appear that there are some genes that are better predicted by methylation compared to expression, though not significantly better. 
* What do the results of these scatter plots mean for your analysis? Do you think you will try to incorporate methylation data still? Or was this contingent on finding a subset of cases where methylation data was more predictive compared to gene expression?

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -228,3 +228,58 @@ def label_points(x, y, gene, sig, ax):
...
+preds_expression_df = au.load_preds_to_matrix(preds_dir, sample_info_df,
+                                              training_data='expression')
+print(preds_expression_df.shape)
+preds_expression_df.iloc[:5, :5]
+
+
+# In[11]:
+
+
+sns.set({'figure.figsize': (15, 10)})
```
ajlee21, line 253: * Just to clarify, these confusion matrices are showing how well training on cancer type A (y-axis) can predict cancer type (x-axis)?
* Looks like you have some positive cases that you used to check that there is some gain in information across cancer types. But overall it looks like there isn't much of a gain using different cancer types.
* Looking at the 2 cases where you have cancers that contribute equally to the prediction performance across all cancer types, my intuition says that if these were rare cancers then I would've expected a lower performance. Since the performance is slightly high, this makes me think that there is some common signal in these 2 cancers that is maybe shared across all cancer types...

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -228,3 +228,58 @@ def label_points(x, y, gene, sig, ax):
...
+preds_methylation_df = au.load_preds_to_matrix(preds_dir, sample_info_df,
+                                               training_data='methylation')
+print(preds_methylation_df.shape)
+preds_methylation_df.iloc[:5, :5]
+
+
+# In[13]:
+
+
+sns.set({'figure.figsize': (15, 10)})
```
ajlee21, line 276: Definitely a lot more noise in the methylation data. 

* Looks like methylation data is not as predictive within the same cancer type compared to using expression data. However there does appear to be more cross information gained. Looks like you're working on some clustering analysis, I wonder if those off diagonal entries are from the same tissue type -- is methylation data very tissue specific?

ajlee21: Interesting...so maybe gene expression data is most predictive when you're within the same context, but methylation can be helpful if you are trying to transfer information to a new context!

ajlee21 (approved):
A few more comments

jjc2718:
> * Just curious, is the alpha = 0.001 standard when using human cancer data? It seems very stringent, but perhaps that is because there are many genes found so this threshold needs to be very strict.

Not really, 0.05 would probably be more standard AFAIK. I originally just used a more stringent threshold since some of the sample sizes we're working with are fairly small, but maybe a better solution would be to run more cross-validation replicates, to give us more confidence that we're not seeing too many false positives.

> * Very minor, do you want to add FILTER_CANCER_TYPES to your cfg file?

This is more of a parameter to the notebook than a general experiment parameter (e.g. we would set this if we want to focus on visualizing results for a cancer type or a few cancer types that we're interested in). So I think it's probably best to keep it here.

> * So your boxplot is showing how well gene expression vs methylation data can predict if a sample is cancer vs normal?

Almost - we're not using any of the normal/control samples from TCGA at the moment, we're just predicting whether a tumor sample is from one cancer type or the rest of the cancer types. So if we're interested in BRCA (for example), all the BRCA samples would get a 1 label and all the other samples (from other cancer types) would get a 0 label. Then we split the dataset (stratified by cancer type) and try to distinguish BRCA from all other cancer types.

> * OVerall it looks like methylation data doesn't perform as well compared to expression data and there is a lot more variance. Though there are some where the prediction looks to be equally high using both expression and methylation, I wonder what is happening for these cases. Looking at your scatter plots it looks like there are only a few cancer types where expression data is significantly more predictive compared to methylation data.

Yeah, I agree with this in general. I think we expected that there would be a decent amount of redundancy in the two data types (since methylation patterns often correlate with expression patterns), but we expected to see some cancer types and/or genes where methylation was a better predictor than expression (e.g. where methylation is being directly affected by the mutation or cancer in question, which would make it a more direct readout of the cancer state than gene expression), and that doesn't really seem to be the case at all.

> * Looks like you did a similar analysis comparing performance using expression vs methylation on the task of predicting mutation or not in plot_mutation_results.ipynb. Might be helpful to add a sentence at the header to describe the task here.

Right, they're very similar scripts. I'll add a header to both.

> According to your scatter plot it does appear that there are some genes that are better predicted by methylation compared to expression, though not significantly better.

Correct - there are a few genes where we see very weak improvement in the methylation data compared to expression, but no examples where methylation is clearly better.

> * What do the results of these scatter plots mean for your analysis? Do you think you will try to incorporate methylation data still? Or was this contingent on finding a subset of cases where methylation data was more predictive compared to gene expression?

Well, I think (and Casey seems to agree) that this could be an interesting finding in its own right (i.e. that methylation is almost never a better readout of the downstream effects of a driver mutation than gene expression), since there are lots of papers that suggest using methylation data for other cancer-related prediction tasks (survival analysis, subtype stratification, etc). We still need to do a lot more work to confirm this, though. In particular, we could have made different choices in preprocessing the methylation data, so we'll have to verify that our results are robust to different preprocessing pipelines.

I also think the question of whether or not methylation data adds any additional predictive ability *on top of* expression data is an interesting one: i.e. can we combine the two data modalities to get better predictions, or are they completely redundant? Probably not something I'll get to in the near future, but I'm interested in looking at this down the road

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: > * Just to clarify, these confusion matrices are showing how well training on cancer type A (y-axis) can predict cancer type (x-axis)?

Exactly. So like I mentioned in my other comment, we're doing one-vs-rest classification of the cancer types. The cancer type on the y-axis is what the model is trained to predict (i.e. the positive label in the training set), and the cancer type on the x-axis is what we're evaluating on (i.e. the true label in the test set). So a high off-diagonal value in the heatmap indicates that a model trained to distinguish the y-axis cancer type from all the others often "confuses" it for the x-axis cancer type (i.e. they have similar gene expression profiles).

> * Looks like you have some positive cases that you used to check that there is some gain in information across cancer types. But overall it looks like there isn't much of a gain using different cancer types.

Right; or stated slightly differently, the models are pretty accurate overall (high on-diagonal values and mostly low off-diagonal values), and the "inaccuracies" (high off-diagonal values) we see aren't surprising based on our knowledge of relationships between cancer types. We expected this result, based on Greg's past work which showed that cancer type prediction is a fairly easy problem for the most part.

> * Looking at the 2 cases where you have cancers that contribute equally to the prediction performance across all cancer types, my intuition says that if these were rare cancers then I would've expected a lower performance. Since the performance is slightly high, this makes me think that there is some common signal in these 2 cancers that is maybe shared across all cancer types...

Yeah, could be! I'm running some experiments now to subsample cancer types (so sample sizes will be uniform across cancers), so if we still see these effects we'll know it's not sample size related.

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: > * Looks like methylation data is not as predictive within the same cancer type compared to using expression data. However there does appear to be more cross information gained. Looks like you're working on some clustering analysis, I wonder if those off diagonal entries are from the same tissue type -- is methylation data very tissue specific?

Yeah, I'll have to look into it! I think it does tend to be tissue-specific, but I'm not sure (and I'm not sure if I'd expect it to be more or less tissue-specific than expression).

> Interesting...so maybe gene expression data is most predictive when you're within the same context, but methylation can be helpful if you are trying to transfer information to a new context!

We were thinking along these lines too...stay tuned!

PR #13, opened 2021-01-11
Subsample cancer type prediction data + cluster confusion matrices 

jjc2718:
There are two main additions (both fairly small in terms of code changes) in this PR:

* In my last PR, we were confused about some "horizontal lines" in the confusion matrix for cancer type prediction on gene expression data:

image

We hypothesized that this was due to label imbalance between different sized cancer types. To test this, we subsampled each cancer type to the size of the smallest cancer type (~50 samples), and repeated the same experiment. This seems to work reasonably well:

image

* I added some clustering analysis of confusion matrix results (see `01_classify_stratified/plot_cancer_type_results.ipynb`). We're hoping to use these results for some downstream analyses, and they look fairly reasonable to me, no huge surprises.


01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -283,3 +285,136 @@ def label_points(x, y, gene, sig, ax):
 plt.title('Cancer type confusion matrix, methylation data', size=14, pad=14)
 plt.tight_layout()
 
+
+# How to read the confusion matrices:
+# 
+# * Remember, we're doing one-vs-rest classification of the cancer types. The cancer type on the y-axis is what the model is trained to predict (i.e. the positive label in the training set), and the cancer type on the x-axis is what we're evaluating on (i.e. the true label in the test set).
```
ajlee21, line 291: Just to clarify, the first column in the bottom row is showing the accuracy of a model to distinguish ACC vs other cancer types, where the model was trained to distinguish UVM from other types of cancers? So a low performance value indicates that the methylation for UVM isn't able to predict ACC. So there isn't shared methylation information between these two cancer types?

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -283,3 +285,136 @@ def label_points(x, y, gene, sig, ax):
...
+# 
+# In the resulting heatmaps, we can observe a few things:
+# 
+# * The models are accurate overall (high on-diagonal values and mostly low off-diagonal values), and the "inaccuracies" (high off-diagonal values) we see aren't surprising based on our knowledge of relationships between cancer types.
+# 
+# * Cancer type predictions using methylation data are noisier and not quite as accurate compared to gene expression data.
+
+# ## Hierarchical clustering analysis
+# 
+# We want to see if these confusion matrices lead to reasonable "cancer groups" (i.e. are cancer types that we would expect to be similar close by in the "similarity matrix" given by model confusion?). To do this, we'll just do a super simple agglomerative hierarchical clustering.
```
ajlee21, line 303: Think there might be a typo in the parenthesis

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -283,3 +285,136 @@ def label_points(x, y, gene, sig, ax):
...
+    
+# fit agglomerative clustering model to confusion (similarity) matrix
+model = AgglomerativeClustering(distance_threshold=0,
+                                n_clusters=None,
+                                affinity='precomputed',
+                                linkage='complete')
+# takes a distance matrix (1 - similarity)
+model = model.fit(1 - preds_expression_df.values)
+
+plot_dendrogram(model, labels=preds_expression_df.index.values)
```
ajlee21, line 337: Whats your takeaway on the grouping? Cancers that group together are those that have similar methylation or expression performance patterns, meaning that the model trained on UVM has a similar performance compared to the model trained on SKCM? If the models perform similarly, what does that mean in terms of how you can use this information?

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -283,3 +285,136 @@ def label_points(x, y, gene, sig, ax):
...
+
+plot_dendrogram(model, labels=preds_methylation_df.index.values)
+plt.title('Dendrogram for complete-linkage hierarchical clustering, methylation data',
+          size=14, pad=10)
+plt.xlabel('Cancer type', size=14)
+
+
+# ## Confusion matrix and clustering with subsampled cancer types
+# 
+# In the expression confusion matrix above, there are some strange "horizontal bars" (e.g. for PAAD and KICH) that we think are an artifact of label imbalance between cancer types of different sizes. To test this, we subsampled each cancer type down to the size of the smallest cancer type (~50 samples) and ran the same classification experiments, taking the average over 5 different random seeds to account for variability in the subsampling.
```
ajlee21, line 362: Can you elaborate a bit more on how the label imbalance could cause these horizontal lines? So if cancer A had 50 samples vs all other cancers have many more samples, then you think the model would learn to classify all samples as negative . Wouldn't that produce variable performance across different test cancers because different test cancers will have different numbers of samples?

01_classify_stratified/nbconverted/plot_cancer_type_results.py
```diff
@@ -283,3 +285,136 @@ def label_points(x, y, gene, sig, ax):
...
+
+plot_dendrogram(model, labels=preds_expression_ss_df.index.values)
+plt.title('Dendrogram for complete-linkage hierarchical clustering, subsampled expression data',
+          size=14, pad=10)
+plt.xlabel('Cancer type', size=14)
+
+
+# The confusion matrix and clusters from the subsampled expression data look much more reasonable to us: the subsampling seems to get rid of the horizontal 'bars' from the non-subsampled analysis (fixes label imbalances), and most things that we expect to be clustered together are together (GBM/LGG, READ/COAD, SKCM/UVM, UCEC/UCS, DLBC/LAML). There are some idiosyncracies (KICH isn't with KIRC/KIRP, all are kidney cancers) but overall this seems like a reasonable starting point.
+# 
+# I also tried "smallest" and "average" linkage clustering - both give somewhat reasonable results, but "complete" linkage clustering gives the clearest separation between related cancer types, at least to my (relatively untrained) eye.
```
ajlee21, line 410: What's the difference between these different linkings?

ajlee21 (approved):
The code all looks good to me, just had some questions. Looks very interesting!

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: No typo (that I know of), but the wording here is def awkward. I'll rephrase to hopefully be a bit clearer.

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: Close - it's showing the accuracy of a model trained to distinguish ACC vs. other cancer types on test samples whose true label is UVM. If the model was perfectly accurate (always classified ACC samples and other samples correctly), there would be 1 on the diagonal and 0 everywhere else, since the model would only predict UVM samples as 1 and everything else as 0.

So a low value indicates that it's rare for the model to "confuse" UVM with ACC, suggesting that their methylation patterns probably aren't very similar.

Does that make more sense? It's a bit of a counterintuitive way to define similarity between cancer types, since we're using supervised classifiers to do it (normally you'd just use some sort of unsupervised distance metric). But we had the data already, so we decided it's a good place to start.

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: Hopefully my first comment on what the matrix means makes this a bit clearer. Cancers that group together are ones the model tends to mistake for one another (UVM and SKCM are both types of melanoma, so it makes sense that they have similar expression and methylation, and a model trained to differentiate one from all other cancer types will sometimes mistakenly identify the other).

So it's not just about the model performance in general, but about the specific instances where the model makes mistakes, which can tell us something (in a bit of a roundabout way) about which cancers are similar.

We're going to try some more direct methods of calculating "distance" between cancer types in the future, so maybe that analysis will be a bit more intuitive than this one. It's not typical to use confusion matrices in this way, so it's totally understandable that things are unclear to you.

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: Agglomerative hierarchical clustering essentially works by iteratively merging elements in a dataset (in this case, cancer types) based on some similarity or distance measure (in this case, confusion matrix probability) until a certain number of clusters are formed.

When you're comparing two individual elements/cancer types it's fairly obvious how this would work (just combine the most similar cancer types based on entries in the similarity matrix, then repeat with the next most similar, and so on), but eventually to form larget clusters you have to ask the question: how should we compare two clusters/groups of elements (or one cluster and a single element, etc)?

You can imagine lots of different ways to do this, but probably the most intuitive are:
* "Smallest" linkage clustering: distance between clusters = min distance between individual elements
* "Average" linkage clustering: distance between clusters = average distance between individual elements
* "Complete" linkage clustering: distance between clusters = max distance between individual elements

For whatever reason, complete linkage seems to work the best here. I don't have a great rationale for why that is, but my intuition is that it maximizes the effect of the worst-case comparisons -- that is, if there are a few very similar cancer types and one dissimilar cancer type, this will have a large complete linkage distance.

I don't think it would make a huge difference if I used average linkage, but smallest linkage does seem to give weird clusters in some cases.

01_classify_stratified/nbconverted/plot_cancer_type_results.py
jjc2718: The performance might be variable, but what I'm plotting is the probability of the positive class, which I think would be (almost) the same across different test cancers in the case you're describing.

Truthfully I don't think the reason it works has totally clicked for me yet (although your questions are helping me make sense of it!), but in practice the subsampling does seem to fix the problem. Not a very satisfying answer, I know...

01_classify_stratified/nbconverted/plot_cancer_type_results.py
ajlee21: So a low value means that your UVM-trained model could not successfully distinguish between ACC vs other cancers samples? Suggesting that the methylation patterns are not similar?

01_classify_stratified/nbconverted/plot_cancer_type_results.py
ajlee21: I think that's how it goes sometimes...sometimes you have to find a solution and then back track why it works:) Something I'm also learning

01_classify_stratified/nbconverted/plot_cancer_type_results.py
ajlee21: That logic makes sense to me

PR #14, opened 2021-01-28
Process 450K methylation data and filter to overlapping samples

jjc2718:
This PR does a few different things (sorry for the disjointed change set), most of which are small-ish code changes:

* Implement filtering and imputation for methylation data (rather than just filtering out all probes with any NA values)
* Rerun mutation classification on filtered/imputed methylation data (results don't change)
* Download and preprocess 450K methylation data. What we have been using is 27K methylation data (27,000 CpG probes), but TCGA also has data with 450,000 probes (for slightly fewer samples) that we want to compare against as well.
* Add code to filter to the samples that all of the data types have in common, so that we're comparing apples to apples

The main purpose is to set up for the next set of experiments I'm going to run, comparing classification on compressed (PCA) representations of the data. This will allow us to use the 450K probe methylation data, since running logistic regression on this many features without any preprocessing is challenging computationally.


00_download_data/1C_preprocess_450k_methylation_data.ipynb
```diff
@@ -0,0 +1,471 @@
...
+    "## Preprocess pan-cancer 450K methylation data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Load the downloaded data and curate sample IDs.\n",
+    "\n",
+    "TODO: look at cancer type composition of represented samples (are any missing?)"
```
ben-heil, line 16: There's still a TODO here, you may want to convert it to a Github issue so it doesn't get lost

mpmp/config.py
```diff
@@ -6,18 +6,42 @@
...
+# locations of processed multimodal data files
+rnaseq_data = data_dir / 'tcga_expression_matrix_processed.tsv.gz'
+# use this as default methylation data for now
+# this is 27K methylation data, 450K is too large to use unprocessed
+methylation_data = data_dir / 'methylation_preprocessed' / 'methylation_processed_n10_i5.tsv.gz'
+data_types = {
+    'expression': rnaseq_data,
+    'methylation': methylation_data,
+}
+use_only_cross_data_samples = True
```
ben-heil, line 24:
```suggestion
# If true, use only the samples present in both the 27k and 450k methylation datsets
use_only_cross_data_samples = True
```

mpmp/utilities/classify_utilities.py
```diff
@@ -115,7 +115,7 @@ def run_cv_stratified(data_model,
                 max_iter=cfg.max_iter
             )
         except ValueError as e:
-            if 'Only one class' in str(e):
+            if ('Only one class' in str(e)) or ('got 1 class' in str(e)):
```
ben-heil, line 118: Is this catching errors thrown by sklearn and rewrapping them so you can filter them later?

ben-heil (approved):
Looks good! Sorry that it took so long to review, I didn't see the notification on Friday

00_download_data/1C_preprocess_450k_methylation_data.ipynb
jjc2718: Good idea! Created #16

mpmp/utilities/classify_utilities.py
jjc2718: Yeah, exactly. I don't want this error to completely stop the script that calls this function (which it does if it isn't caught); I just want to skip the gene that this happens for, and move on to the next gene after logging the error.

I'm not totally sure why there are two different types of single-class errors that get thrown in this sklearn function, but I'm fine with treating them the same here.

PR #17, opened 2021-02-03
Compressed classification/450K methylation pilot study

jjc2718:
**Changes in this PR:**

* Add script to predict mutation status from compressed (PCA transformed) data
* Run script for 3 data types (expression, 27K methylation, 450K methylation) across 3 PCA dimensions (100, 1000, 5000)
* Plot results, comparing data types against baseline with shuffled labels and against each other

**Results/takeaways:**

Mutation prediction based on gene expression data still largely performs better than prediction based on methylation data, but using 450K methylation data definitely brings performance closer to the expression data (and makes it slightly better in a few cases), as seen here:

image

Median prediction performance is higher for PCA dimensions 1000 and 5000 in expression data:

image

These results lead me to believe that we should probably use the 450K methylation data in the future (in most cases it seems to be a fairly clear improvement on the 27K data). We'll have to get creative about how to use it due to the high dimensionality, but based on these results it's probably worthwhile to think about.


02_classify_compressed/run_mutation_compressed.py
```diff
@@ -0,0 +1,228 @@
...
+                                            model_options.training_data,
+                                            sample_info_df,
+                                            model_options.num_folds,
+                                            shuffle_labels,
+                                            standardize_columns=False)
+            except NoTrainSamplesError:
+                if io_args.verbose:
+                    print('Skipping due to no train samples: gene {}'.format(
+                        gene), file=sys.stderr)
+                mutation_log_df = fu.generate_log_df(
```
ajlee21, line 204: This is cool. I'll need to keep this in mind. I know Milton mentioned this as something to consider for future projects

02_classify_compressed/nbconverted/plot_mutation_compressed_results.py
```diff
@@ -0,0 +1,303 @@
...
+                    lim=5)
+
+
+# In[6]:
+
+
+sns.set({'figure.figsize': (18, 6)})
+fig, axarr = plt.subplots(1, 2)
+
+# plot mean performance over all genes in Vogelstein dataset
```
ajlee21, line 144: When you're farther along, might be helpful to pull this into a function since I see you're making similar plots in other ntoebooks

ajlee21 (approved):
All the code looks good to me, just had a couple of comments about the findings:

1. Just to clarify, you say that 450K methylation data improves performance to be similar to expression because you're observing that the genes are roughly in the middle of the volcano plots in the middle column?
2. So 27K vs 450K methylation data appear to be roughly equally predictive based on your volcano plots, and yet when you compare each of the methylation datasets against the expression datasets you're seeing a difference. I wonder what could explain this.
3. I may have missed this from your presentation today, but how many features do you have for your datasets? I'm curious how much reduction 5K PCs is. It does look like you're seeing some genes pulling away with more PCs. Looks like you're getting a particular gain in ERBB2 gene with more PCs for expression data. But your SETD2 seems to disappear with more PCs and its a gene that was slightly more predictive using methylation data. Maybe some things to consider after your committee meeting. I'd be curious to know.

02_classify_compressed/nbconverted/plot_mutation_compressed_results.py
jjc2718: Good idea! I'll create an issue

jjc2718:
Responses to your questions below:

> 1. Just to clarify, you say that 450K methylation data improves performance to be similar to expression because you're observing that the genes are roughly in the middle of the volcano plots in the middle column?

Yep, exactly (especially compared to the 27K methylation plots in the left column, where there are quite a few genes to the left of the line for 0). The box plots also support this, particularly the ones that are looking at the median performance when filtered to significant genes only (right plot).

> 2. So 27K vs 450K methylation data appear to be roughly equally predictive based on your volcano plots

I don't think this is entirely true - for 100 PCs (top right plot) maybe, but for 1000/5000 PCs (bottom right) it definitely looks to me like there are more points toward the right of the midline, indicating better performance using 450K data. Most of them don't rise to the level of statistical significance (above dotted blue line), but the signal's still there somewhat IMO.

> and yet when you compare each of the methylation datasets against the expression datasets you're seeing a difference. I wonder what could explain this.

Yeah, it's definitely a stronger effect in the methylation vs. expression comparisons. I'll have to think about this, I don't have a great answer off the top of my head.

> 3. I may have missed this from your presentation today, but how many features do you have for your datasets? I'm curious how much reduction 5K PCs is. It does look like you're seeing some genes pulling away with more PCs. Looks like you're getting a particular gain in ERBB2 gene with more PCs for expression data. But your SETD2 seems to disappear with more PCs and its a gene that was slightly more predictive using methylation data. Maybe some things to consider after your committee meeting. I'd be curious to know.

* 27K is about 20,000 features after filtering/imputation
* 450K is about 300,000 features after filtering/imputation
* Gene expression is about 15,000 features after filtering NA values (didn't do any imputation)

The overlap of all 3 datasets is about 10,000 samples, so they're all "wide" datasets (more features than samples).

I'm not sure how many of these features are actually selected by the elastic net, though - I might take a look at that and see if it varies substantially between the data types.

And yeah, Casey and I were talking about the ERBB2 and SETD2 examples in our meeting this week. It's definitely plausible that some genes affect lots of pathways in complex ways so their prediction gets better gradually as more PCs are added, and other genes have more simple/straightforward effects and so they "saturate" more quickly as PCs are added. Something we're for sure interested in thinking more about, but probably won't dig deeper until after my committee meeting.

PR #19, opened 2021-02-08
PCA variance explained + elastic net nonzero feature distributions

jjc2718:
**Changes in this PR:**

This answers some of the questions that @nrosed brought up during my practice committee meeting talk last week:
* Look at variance explained by truncated/"thin" PCA dimension reduction, for different data types (expression/27k methylation/450k methylation)
* Look at distributions of nonzero features across different data types (only for expression and 27k methylation, I haven't run classifiers on untransformed 450k data yet)

**Results:**

The trend in variance explained looks essentially the same across the data types: 100 PCs explain around 60-70% of variance, 1000 PCs explain around 80-85% of variance, and 5000 PCs captures around 95-98% of variance. I wouldn't expect the extra 5% of variance to be particularly predictive of most mutations, so taking 5000 PCs as our maximum should be fine (at least I think so).

Here's an example for the 450K methylation data - you can see the other data types in the notebooks.

image

In the nonzero coefficient analysis, the gene expression models tend to have more nonzero coefficients than the 27k methylation models, despite starting with fewer features (~15,000 for gene expression vs. ~20,000 for methylation). This is particularly true if we filter to examples where the target gene is well-predicted.

image


nrosed:
Variance plots looks good, it matches what I would expect for expression data. For the nonzero coefficient analysis, what is the null hypothesis? Also is this analysis for a single cancer type? Is this for all mutations?

jjc2718:
Ahh, good questions - answers below:

> is this analysis for a single cancer type? Is this for all mutations? 

This analysis is for models that are stratified by cancer type (train/test sets have equal proportions of all cancer types). I haven't looked at this in the case where we're holding out a single cancer type yet, that might be an interesting next step.

The distribution combines models for all genes in the Vogelstein & Kinzler gene set (gene list [here](https://github.com/greenelab/pancancer/blob/master/data/vogelstein_cancergenes.tsv), original reference [here](https://www.nature.com/articles/nm1087)).

> For the nonzero coefficient analysis, what is the null hypothesis?

In addition to looking at all genes together, I wanted to see if there was a difference in the number of nonzero coefficients for genes where we can train a "good" predictor vs. genes where we can't. The way I've been distinguishing between these cases in some of my past experiments is by running cross-validation a few times (2 replicates x 4 folds), then doing a t-test to compare with a model where the labels were shuffled. So, the null hypothesis is that the distribution of cross-validation results (AUPR values) is the same with the true mutation labels as with shuffled mutation labels, and the alternative hypothesis is that it's different (in this case, the model with true labels is always better, although I am using a two-tailed test).

So looking at the results, it does look to me like there's a bit of a difference: for genes that we can predict "well", elastic net generally seems to select more features than it does for genes where there's little or no predictive signal. This isn't too surprising, mostly just a sanity check.

nrosed:
Aha! Ok, cool I get it now. Yes it makes sense to me, but I don't have a strong intuition why a predictive model would have more features than a non-predictive one. 
Also, how are you transforming your expression/ methylation data? Do they use the same transform, just curious for myself.

nrosed (approved):

jjc2718:
We've been standardizing expression data (TCGA provides RPKM values, then we standardize them for each gene independently), and currently we're just using the methylation array beta values provided by TCGA with no standardization/normalization.

We tried standardizing the methylation beta values too in some previous experiments, and it didn't seem to make any difference as far as the results, so we stuck with the simpler approach.

00_download_data/nbconverted/1A_preprocess_expression_data.py
```diff
@@ -234,13 +234,36 @@
...
+
+sns.set({'figure.figsize': (15, 4)})
+fig, axarr = plt.subplots(1, 3)
+
+for ix, ve in enumerate(var_exp_list):
+    sns.lineplot(x=range(len(ve)), y=ve, ax=axarr[ix])
+    axarr[ix].set_title('{} PCs, total variance explained: {:.4f}'.format(
+        n_pcs_list[ix], sum(ve, 0)))
+    axarr[ix].set_xlabel('# of PCs')
+    if ix == 0:
```
ben-heil, line 265: It doesn't really matter unless this is going to end up as a publication figure, but IMO variance explained graphs make more sense if you plot the cumulative variance explained instead of the individual variance per PC

01_classify_stratified/nbconverted/coefficient_analysis.py
```diff
@@ -0,0 +1,134 @@
...
+axarr[0].legend(h, l, title=r'Reject $H_0$')
+
+sns.histplot(data=coefs_sig_df[coefs_sig_df.training_data == 'methylation'], 
+             x='nz_coefs', hue='reject_null', element='step', ax=axarr[1])
+axarr[1].set_xlabel('Number of nonzero coefficients')
+axarr[1].set_xlim(0, x_max)
+axarr[1].set_title('Nonzero coefficient distributions, methylation data')
+h = axarr[1].legend_.legendHandles
+l = [t.get_text() for t in axarr[1].legend_.get_texts()]
+axarr[1].legend(h, l, title=r'Reject $H_0$')
```
ben-heil, line 107: It looks like the distribution of nonzero coefficients is bimodal, do you have an intuition for why that might be? It's especially interesting that the distribution of positive and negative expression models' coefficients look like they might be bimodal and unimodal respectively

ben-heil (approved):
Looks great!

01_classify_stratified/nbconverted/coefficient_analysis.py
jjc2718: I don't have a great intuition of why this would be, off the top of my head. I'm not surprised that the distribution has a fairly high variance (some mutations might have a very strong effect on a single pathway, others might affect many different pathways), but I'm not sure why that would be bimodal, since I'd also expect to see some cases in between.

Definitely interested in looking into this more in the future - would be interesting to see how robust this bimodality is (e.g. across cancer types), or if it persists when we use different feature selection methods.

00_download_data/nbconverted/1A_preprocess_expression_data.py
jjc2718: Good call! Switched to plotting cumulative variance explained.

PR #21, opened 2021-02-15
Classification experiments for 450K methylation data with raw features

jjc2718:
Our previous experiments showed that compressed 450K methylation data was comparable in terms of performance to compressed gene expression data, so our next step is to try raw features (CpG beta values) for 450K methylation data and see how that compares to raw gene expression values.

The main takeaways are in the `compressed_vs_raw_results.ipynb` script. This plot is a fairly representative summary of what we found: 

image

The left plot shows the performance distribution over different PCA dimensionalities (+ raw data) for all genes in our gene set, and the right is the same plot for genes that are significantly better than the random baseline for at least 1 data type. Both plots show that gene expression data with raw features is generally the best performing, and for the methylation data it doesn't seem to matter much whether we use the PCA representation or the raw features.


ajlee21 (approved):
Your code looks good to me, just some conceptual questions as always:

1. I'm surprised that your compression didn't improve your prediction and that you're seeing better prediction using raw data, which I think to be noisier overall. This is making me wonder if there are a few very key genes that are predictive and the PCs are muddling that signal. I wonder what your opinion is about this though. Perhaps this didn't seem surprising to you.
2. Can you clarify what these significant genes are? How many points are represented in that bar plot, is it just a few? If there are very few, perhaps plotting the raw points would be better in this case.
3. I'm not as familiar with methylation data, so I wonder if its worth "looking more at the data" to determine if there are any distinct clustering patterns within CpG values similar to what you already did with the expression data.

jjc2718:
Good questions!

> 1. I'm surprised that your compression didn't improve your prediction and that you're seeing better prediction using raw data, which I think to be noisier overall. This is making me wonder if there are a few very key genes that are predictive and the PCs are muddling that signal. I wonder what your opinion is about this though. Perhaps this didn't seem surprising to you.

I wasn't surprised in the sense that this is similar to what Greg found in his BioBombe work: if you look at [Figure S13 here](https://figshare.com/articles/journal_contribution/Additional_file_1_of_Compressing_gene_expression_data_using_multiple_latent_space_dimensionalities_learns_complementary_biological_representations/12285494), the blue dotted line is raw gene expression data and the points are compressed data. So there are a few cases there, too (PTEN, KRAS, ARID1A) where using the raw expression data performed better than any of the compressed datasets.

I'm not sure why this is, though. What you're seeing seems plausible - maybe [sparse PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.SparsePCA.html) or some other sparse dimensionality reduction would give us better performance. The other possibility is that the predictive signal for certain mutations isn't well represented in the directions of highest variance in the gene set (or in other words, PCA is an unsupervised algorithm that we're using for a supervised problem). We could try a "supervised dimension reduction" approach (e.g. [LDA](https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html)) to see if that performs comparably, although I'm not sure how that would handle such a high dimensional dataset.

Created #22 to think about these questions, I think they're interesting ones.

> 2. Can you clarify what these significant genes are? How many points are represented in that bar plot, is it just a few? If there are very few, perhaps plotting the raw points would be better in this case.

There are 41 significant genes (out of a total of about 100). If I add a plot of the raw points it looks like this:

image

The bar plot seems like a fairly accurate representation of these distributions in this case to me (in other words, they don't seem terribly undersampled or anything like that).

> 3. I'm not as familiar with methylation data, so I wonder if its worth "looking more at the data" to determine if there are any distinct clustering patterns within CpG values similar to what you already did with the expression data.

Yeah - I haven't done any exploration of the 450K methylation dataset yet just because it's so high dimensional. I can probably PCA it down to 2 dimensions, but I doubt UMAP will scale as well to 100K+ input dimensions.

I'll have to think of some reasonably efficient ways to do this, but you're right, it would be interesting to do more exploration in the future.

PR #23, opened 2021-02-16
RPPA data preprocessing and sample analysis

### LINKED ISSUE #16: Look at cancer type of filtered samples in methylation data
See:  https://github.com/greenelab/mpmp/pull/14#discussion_r567829485

Would be helpful to know if any cancer types are more or less represented after filtering, since we're doing stratified classification this could throw things off somewhat.

jjc2718:
Closes #16. This PR does the following:

* Download and preprocess RPPA data
* Look slightly closer at sample overlap between data types

RPPA ==  "reverse phase protein array", a type of proteomics technology

The second point above is especially important now that we're adding the RPPA data. We have to use a much smaller set of samples that have data for all of the data types, so we want to know more about which samples we're getting rid of.

Unfortunately, the samples that don't have all data types seem to be roughly uniform/close to uniform across cancer types. This means we can't just run the analyses for the cancer types that have all data modalities.

image


00_download_data/nbconverted/1B_preprocess_27k_methylation_data.py
```diff
@@ -217,9 +218,77 @@ def impute_leq(methylation_df, n_na):
...
+
+# In[11]:
+
+
+# commit from https://github.com/cognoma/cancer-data/
+sample_commit = 'da832c5edc1ca4d3f665b038d15b19fced724f4c'
+url = 'https://raw.githubusercontent.com/cognoma/cancer-data/{}/mapping/tcga_cancertype_codes.csv'.format(sample_commit)
+cancer_types_df = pd.read_csv(url,
+                              dtype='str',
+                              keep_default_na=False)
```
ben-heil, line 233: Referencing data from a specific commit is definitely a good idea. Does it make sense to have the commit live in the config file in case you want to change versions later?

ben-heil: Especially since you use the commit in multiple notebooks

mpmp/utilities/tcga_utilities.py
```diff
@@ -375,3 +375,56 @@ def filter_to_cross_data_samples(X_df,
...
+
+    # extract sample type in the order of the gene expression matrix
+    tcga_id = pd.DataFrame(tcga_df.index)
+
+    # extract the last two digits of the barcode and recode sample-type
+    tcga_id = tcga_id.assign(sample_type = tcga_id.sample_id.str[-2:])
+    tcga_id.sample_type = tcga_id.sample_type.replace(sampletype_codes_dict)
+
+    # extract the first two ID numbers after `TCGA-` and recode cancer-type
+    tcga_id = tcga_id.assign(cancer_type = tcga_id.sample_id.str[5:7])
```
ben-heil, line 407: Is there a guarantee that these IDs will always work the same way, or should you split on TCGA and take the first two characters of the second string?

00_download_data/nbconverted/2_sample_intersection.py
```diff
@@ -0,0 +1,227 @@
...
+print('\n'.join(['{}\t{}'.format(n, len(v)) for n, v in sample_lists.items()]))
+
+
+# We'll start by just counting the number of samples that have data for each of the relevant data types.
+
+# In[4]:
+
+
+# do venn diagram
+from venn import venn
```
ben-heil, line 44: You may want to move this import to the top since you end up importing it multiple times

ben-heil (approved):
You managed to find the one uniform distribution in all of biology, too bad it's working against you

00_download_data/nbconverted/1B_preprocess_27k_methylation_data.py
jjc2718: Yeah, good catch - I copy/pasted this code but the config file seems like a much better idea

mpmp/utilities/tcga_utilities.py
jjc2718: Good idea. I doubt the TCGA ID format will change, but what you're suggesting seems safer/more robust.

PR #24, opened 2021-02-22
Add RPPA data to classification experiments

jjc2718:
This PR adds RPPA data to the other datasets that we've run mutation classification experiments on (gene expression, 27K methylation, 450K methylation).

Main results are in `01_classify_stratified/plot_mutation_results.ipynb`. This figure is probably the most important:

image

Genes on the left of each of these plots have better performance in expression data. So, this shows that using RPPA data for prediction results in the same general trend as the methylation datasets: no substantial improvement in performance for any of the ~100 genes, and substantially worse performance than gene expression data for ~15 of them.


01_classify_stratified/nbconverted/plot_mutation_results.py
```diff
@@ -37,183 +37,189 @@
...
-                                           metric='aupr',
-                                           correction=True,
-                                           correction_method='fdr_bh',
-                                           correction_alpha=SIG_ALPHA,
-                                           verbose=True)
-expression_results_df.rename(columns={'identifier': 'gene'}, inplace=True)
-expression_results_df.sort_values(by='p_value').head(n=10)
+# load compressed data for me_27k and me_450k
+compressed_results_df = au.load_compressed_prediction_results(results_dir, 'gene')
+print(compressed_results_df.seed.unique())
```
ben-heil, line 58: Is there a reason you print out the seed and dimensionality each time?

ben-heil (approved):
Yet another great day to be studying gene expression :)

01_classify_stratified/nbconverted/plot_mutation_results.py
jjc2718: This was mainly for debugging, but I think it's useful to keep for now since I'm combining results from uncompressed and compressed data in a bit of an ad hoc way at the moment (using compressed data for methylation datasets and uncompressed data for other data types). These checks make sure I'm loading the right data (i.e. two replicates for all data types, PCA n=5000, etc) to make a fair comparison between data types.

I'll add a note to this script explaining what I'm checking and why, and I'll create an issue to clean this up in the future. I'm planning to re-organize these scripts so that the order they run in makes more sense (the compression analysis will come first, which will justify using compressed methylation data for subsequent analyses) but that'll be a future PR.

PR #29, opened 2021-03-05
Add automated testing and other cleanup

### LINKED ISSUE #26: Add sample overlap data types to model options
These should be saved with experiment metadata

### LINKED ISSUE #27: Tests
See title

Specifically, model regression tests (like the ones in pancancer-evaluation repo) would be a good place to start

jjc2718:
**PR description:**

This PR addresses a few small outstanding issues:

* Add tests (essentially the same ones as in my other `greenelab/pancancer-evaluation` repo) and set up GitHub actions to run them for new PRs. Closes #27.

* Add the data types that are being used to decide which samples to use to the metadata being saved in the output directory for each experiment. This addresses the first half of #26.

* Clean up a hack I was using to use non-compressed data to select the sample set for certain compressed data experiments. Closes #26.

No new results in this PR, but this should make it easier to add new functionality in the future and be more confident that nothing is breaking.

**Files to review:**

All the large `.tsv` files can safely be ignored, they need to be tracked in git so the tests can run but they don't require any review. Most of the testing code was copied over from the `pancancer-evaluation` repo, so it doesn't necessarily need to be reviewed closely, but feel free to spend as much or as little time as you want looking at it.


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,26 @@
...
+                python-version: 3.7
+
+            - name: Setup conda environment
+              run: |
+                $CONDA/bin/conda env update --file environment.yml --name base
+                $CONDA/bin/pip install -e .
+
+            - name: Run tests
+              run: |
+                $CONDA/bin/pytest tests/
```
ben-heil, line 26: This works (and I think is what I use). If you have a need for more complex uses of Conda (or dependency caching) in Github Actions in the future I recommend https://github.com/marketplace/actions/setup-miniconda

mpmp/scripts/generate_test_data.py
```diff
@@ -0,0 +1,58 @@
...
+        np.savetxt(output_file, metrics_df['auroc'].values)
+
+
+if __name__ == '__main__':
+    p = argparse.ArgumentParser()
+    p.add_argument('--verbose', action='store_true')
+    args = p.parse_args()
+
+    # NOTE this will overwrite existing results files
+    # (could eventually add a warning for this?)
```
ben-heil, line 53: Could you also just have separate results files for testing and analyses? You may already do this, I'm not sure

mpmp/utilities/tcga_utilities.py
```diff
@@ -321,34 +321,40 @@ def subsample_to_smallest_cancer_type(X_df,
...
 
-    if debug:
-        data_types = cfg.subsampled_data_types
-    elif compressed_data:
-        data_types = cfg.compressed_data_types
-    else:
-        data_types = cfg.data_types
+    data_types = get_overlap_data_types(use_subsampled, compressed_data_only)
 
+    # TODO this may take some time to load, so we could cache it somewhere
```
ben-heil, line 347: Unclear what 'this' is referring to, but if this is a function you call frequently you could use something like [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) to do caching pretty easily

mpmp/test_config.py
```diff
@@ -0,0 +1,22 @@
...
+Configuration information for tests
+
+"""
+import pathlib
+
+repo_root = pathlib.Path(__file__).resolve().parent.parent
+
+# location of data relevant to running tests
+test_data_dir = repo_root / 'tests' / 'data'
+test_pancan_data = test_data_dir / 'pancancer_data_subsampled.pkl'
```
ben-heil, line 11: Just double-checking, is the `pancancer_data_subsampled.pkl` file that you uploaded supposed to be empty?

ben-heil (approved):

mpmp/scripts/generate_test_data.py
jjc2718: Ah, sorry, this comment isn't very clear - the results files this script generates are only used for testing. I'm just noting that if the files already exist, they'll be replaced by this script (although I don't think that's a huge concern since they're tracked by git, so changes can always be reverted).

I'll add some clarification to the comment.

mpmp/test_config.py
jjc2718: It shouldn't be! When I look at the file [here](https://github.com/greenelab/mpmp/blob/f9cc862dd83952b5d33b51baca063021253bb4d2/tests/data/pancancer_data_subsampled.pkl) it looks like it's a ~2 MB binary file - are you seeing something different somewhere?

jjc2718: All set, this file should be correct (as far as I can tell)

Issue #12, opened 2021-01-07
Use Python f-strings instead of .format()


jjc2718:
See: https://github.com/greenelab/mpmp/pull/9#discussion_r552645289

jjc2718:
https://github.com/greenelab/mpmp/pull/32#discussion_r593217709

PR #32, opened 2021-03-11
Tumor purity prediction pilot study

jjc2718:
**PR description:**

Our previous results (predicting mutation status from various -omics data types) suggested that gene expression tends to perform better for this problem than other data types. Cancer cells tend to have very cell type-specific DNA methylation signatures, which tend to be very different from stromal cells (see, e.g. [this paper](https://doi.org/10.1093/bioinformatics/btv370) or [this paper](https://doi.org/10.1038/onc.2017.176)).

So, we wanted to try using tumor purity prediction as a positive control for our methylation preprocessing pipeline, under the assumption that purity should be more predictable from methylation data than from gene expression. We're starting by binarizing purity values into above/below median and doing classification.

Methylation does seem to perform better for this problem, especially for the 450K probe dataset:

image

**Files to review:**

This looks like a lot of changes, but most of the files are just moved or have very minor edits (I restructured directories a bit). Main new files added include:

* `01_explore_data/purity_distribution.ipynb`
* `02_classify_stratified/run_purity_prediction.py`
* `02_classify_stratified/plot_purity_results.ipynb`

But, of course, feel free to look at other files in as little or much detail as you want.


02_classify_stratified/nbconverted/plot_purity_results.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Plot tumor purity prediction results
+# 
+# Here, we'll visualize the results of our tumor purity prediction. For now, we're trying to predict whether a given sample is above or below the median tumor purity (i.e. classification).
```
miltondp, line 6: Did you try a regression model on the continuous purity value? Just wondering, might not make sense.

02_classify_stratified/nbconverted/plot_purity_results.py
```diff
@@ -0,0 +1,124 @@
...
+plot_df = (
+    results_df[results_df.data_type == 'test']
+      .sort_values(by=['signal', 'training_data'])
+)
+
+sns.boxplot(data=plot_df, x='signal', y='aupr', hue='training_data')
+plt.title('Binarized tumor purity prediction performance, by data type')
+plt.legend(title='Data type')
+plt.xlabel('Signal or shuffled')
+plt.ylabel('AUPR')
```
miltondp, line 61: So this is the PR. What about ROC? Does it make sense to look at that as well?

miltondp: Just curious: is this the average precision? How do you calculate the area under the PR curve?

02_classify_stratified/nbconverted/plot_purity_results.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+# In[5]:
+
+
+# now facet by cancer type
```
miltondp, line 67: I know nothing about cancers. But I wonder why the performance is so low with all data types for PAAD.

00_download_data/nbconverted/0_download_data.py
```diff
@@ -32,7 +32,8 @@
 # In[3]:
 
 
-os.makedirs(cfg.raw_data_dir, exist_ok=True)
+if not os.path.exists(cfg.raw_data_dir):
+    os.makedirs(cfg.raw_data_dir)
```
miltondp, line 36: For a future PR maybe, you can also migrate this to `pathlib`.

Also, did the `exist_ok` failed for some reason that you are changing it here?

00_download_data/nbconverted/0_download_data.py
```diff
@@ -72,14 +73,39 @@
...
 
 # In[7]:
 
 
+me_id, me_filename = manifest_df.loc['methylation_450k'].id, manifest_df.loc['methylation_450k'].filename
+url = 'http://api.gdc.cancer.gov/data/{}'.format(me_id)
+me_filepath = os.path.join(cfg.raw_data_dir, me_filename)
+
+if not os.path.exists(me_filepath):
+    urlretrieve(url, me_filepath)
```
miltondp, line 91: I also used `urlretrieve` in the past, but it failed to download recount2 files when they "moved" the URL. Seems like it doesn't support redirection. I had to change it and use the `requests` package ([here](https://github.com/miltondp/phenoplier/blob/641ba94c6618d3e98e5c5814fc50d61dc624ef39/libs/data/recount2.py#L103), in case you have the same problem in the future).

00_download_data/nbconverted/0_download_data.py
```diff
@@ -72,14 +73,39 @@
...
+else:
+    print('Downloaded data file already exists, skipping download')
+
+
+# In[8]:
+
+
+md5_sum = get_ipython().getoutput('md5sum $me_filepath')
```
miltondp, line 99: Ah, interesting. Python has the `hashlib` package that you might already know, but using `md5sum` seems more direct and easy (although it might not work on Windows or macOS; I think in macOS the command is slightly different).

mpmp/config.py
```diff
@@ -22,8 +22,8 @@
     'expression': expression_sample_info,
     'me_27k': me_27k_sample_info,
     'me_450k': me_450k_sample_info,
-    'rppa': rppa_sample_info,
-    'mut_sigs': mut_sigs_sample_info,
+    # 'rppa': rppa_sample_info,
+    # 'mut_sigs': mut_sigs_sample_info,
```
miltondp, line 26: Do you need to leave this code commented out here? I mention this just in case you don't.

mpmp/utilities/analysis_utilities.py
```diff
@@ -71,6 +72,112 @@ def load_compressed_prediction_results(results_dir, experiment_descriptor):
...
+    results_dir = Path(results_dir)
+    for results_file in results_dir.iterdir():
+        if not results_file.is_file(): continue
+        results_filename = str(results_file.stem)
+        if 'classify' not in results_filename: continue
+        if results_filename[0] == '.': continue
+        id_results_df = pd.read_csv(results_file, sep='\t')
+        if check_compressed_file(results_filename):
+            id_results_df.training_data += '_compressed'
+        results_df = pd.concat((results_df, id_results_df))
```
miltondp, line 99: In case you have performance issues (not sure how big is the data), I think it's more efficient to add all dataframes to a list, and then perform the `pd.concat` at the end.

You don't need to change this. It's just a comment in case you run into performance issues.

mpmp/utilities/file_utilities.py
```diff
@@ -16,56 +16,47 @@ def make_output_dir(experiment_dir, identifier):
...
-        except AttributeError:
-            # no n_dim in model_options => not a compressed model
-            return Path(output_dir,
-                        '{}_s{}_{}{}'.format(training_data,
-                                             seed,
-                                             file_descriptor,
-                                             extension))
+    if len(args) == 0 and len(kwargs) == 0:
+        return Path(output_dir,
+                    '{}{}'.format(file_descriptor, extension))
```
miltondp, line 41: In case you want (and can) use f-strings.

```suggestion
                    f'{file_descriptor}{extension}'
```

miltondp (approved):
Great job, Jake! Looks great to me. I left some comments, all are minor, you can take them or not.

You already got great results to test your hypothesis. I left here this thought in case is useful in the future: instead of using the median, maybe if you binarize by the, say, 25th or 75 percentile, you might get different (or even better) results. Could it be possible that, say, gene expression cannot predict values below/above the median, but only really high purity values?

jjc2718:
Thanks for the great feedback, as always! Response to your question below:

> You already got great results to test your hypothesis. I leave here this thought in case is useful in the future: instead of using the median, maybe if you binarize by the, say, 25th or 75 percentile, you might get different (or even better) results. Could it be possible that, say, gene expression cannot predict values below/above the median, but only really high purity values?

Yep, totally - Casey and I had the same thought. We just did the above/below median prediction as a quick way to see if this was worth pursuing (since I can just plug it in to my existing classification code), but now that we know it works our plan is to predict purity values directly using regression. That'll most likely be my next PR.

I'd be surprised if the results differ substantially (i.e. methylation should still perform better than expression), but it will be interesting to find out!

02_classify_stratified/nbconverted/plot_purity_results.py
jjc2718: Not yet - that's our next experiment!

02_classify_stratified/nbconverted/plot_purity_results.py
jjc2718: > So this is the PR. What about ROC? Does it make sense to look at that as well?

We've generally used PR in our past experiments since it deals better with imbalanced classes (especially when there are very few positively labeled examples; see, e.g. [this paper](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0118432)), but in this case the labels are pretty balanced so ROC would be fine.

In any case, I did look at ROC here too, and it looks essentially the same as the PR results. I'll add those plots to this script as well.

> Just curious: is this the average precision? How do you calculate the area under the PR curve?

Yep, it's average precision, using [the sklearn implementation here](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html).

02_classify_stratified/nbconverted/plot_purity_results.py
jjc2718: I don't know much about pancreatic cancer specifically (and it's debatable whether I know much more than nothing about cancer in general, haha) - but looking at [this page on the TCGA pancreatic cancer study](https://www.cancer.gov/about-nci/organization/ccg/research/structural-genomics/tcga/studied-cancers/pancreatic), it seems like pancreatic tumors tend to have very low purity in general (or in other words, PAAD samples are generally heterogeneous mixtures of stromal cells and cancerous cells, rather than samples of mostly cancerous cells like in many other cancer types).

It's not exactly intuitive to me why this would lead to poor performance: ideally our model would pick up on this and predict lower values for PAAD samples, but maybe the high overall tumor heterogeneity makes it hard to distinguish between samples based just on -omics data? I dunno, I'll have to think it through.

miltondp: Ah, super interesting! I wonder if using other binarization strategies would improve results with PAAD (not sure why it would, but just wondering).

mpmp/config.py
jjc2718: I do - for these experiments I want to use only the overlapping samples between the expression and methylation datasets (if I add the other data types, lots of samples get dropped), and commenting them out here was the easiest way to do that.

I should find a better way to deal with this in the future, though. I'll create an issue to remind myself.

00_download_data/nbconverted/0_download_data.py
jjc2718: Yeah, I remember looking into this before - I think using `hashlib` to get file hashes is a bit more complicated and not super efficient for large files (see, e.g. [this post](https://stackoverflow.com/a/16876405)), so I decided that just piping the file straight to `md5sum` would be easier.

Good to know that it might not work on other platforms, though. I'll make a note in the script.

02_classify_stratified/run_purity_prediction.py
```diff
@@ -0,0 +1,211 @@
...
+                [model_options.training_data, shuffle_labels, 'no_train_samples']
+            )
+        except OneClassError:
+            if io_args.verbose:
+                print('Skipping due to one holdout class', file=sys.stderr)
+            purity_log_df = fu.generate_log_df(
+                log_columns,
+                [model_options.training_data, shuffle_labels, 'one_class']
+            )
+        else:
```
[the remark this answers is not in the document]
jjc2718: Looking at [this post](https://stackoverflow.com/a/855764), I think you're right. Since I'm only catching exceptions that I defined (and that won't happen in the `save_results` function), there shouldn't be any difference here between adding the `else` clause and putting the save code on line 183.

I'll switch to what you're suggesting, since I think it'll be clearer to most people.

00_download_data/nbconverted/0_download_data.py
jjc2718: No, `exist_ok` should be fine - I was just copying old code. I'll update these.

And yeah, I should really update this to `pathlib` in a future PR.

01_explore_data/nbconverted/purity_distribution.py
```diff
@@ -0,0 +1,65 @@
...
+# In[45]:
+
+
+sns.set({'figure.figsize': (8, 5)})
+
+g = sns.kdeplot(data=purity_cancer_df, x='purity', hue='cancer_type', legend=False)
+# TODO: figure out legend
+# g.fig.legend(labels=[], ncol=2)
+plt.gca().axvline(x=purity_df.purity.median(), linestyle='--', color='red')
+plt.xlabel('Tumor purity')
```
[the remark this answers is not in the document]
jjc2718: Will do!

mpmp/utilities/file_utilities.py
jjc2718: I should really update all of my code to use f-strings (I have issue #12 to go through and do that). I'm planning to take a few days to do some code cleanup in the near future, so I'll try to roll this into those changes.

02_classify_stratified/nbconverted/plot_purity_results.py
jjc2718: Yeah, this would be interesting to look at! We'll see if regression improves things, but other binarization strategies would be interesting too.

00_download_data/nbconverted/0_download_data.py
jjc2718: This is good to know! I don't think these will move anytime soon, but I'll keep this in mind for the future.

mpmp/utilities/analysis_utilities.py
jjc2718: Good to keep in mind! This doesn't seem to be too slow, but for future code I'll definitely do what you're suggesting.

PR #34, opened 2021-03-16
Tumor purity regression experiments

jjc2718:
**PR description:**

Following up on #32, we wanted to try predicting exact tumor purity values using elastic net regression, rather than binarizing purity values into above/below median and classifying. This took some engineering work/changes to the code, but the results and interpretation are largely unchanged from the previous PR. Methylation datasets still seem to be more effective than gene expression data for predicting tumor purity values:

image

**Major code changes in this PR:**

* Rearrange the classification code that was previously in `mpmp/utilities/classify_utilities.py` into two separate scripts: one for regression (`mpmp/prediction/regression.py`) and one for classification (`mpmp/prediction/classification.py`). All the code in the `classification.py` script has been previously reviewed; most of the code in the `regression.py` script is new.
* Put code shared between regression and classification (mostly related to cross-validation splitting) into a common file, at `mpmp/prediction/cross_validation.py`. This code has all been reviewed before, it's just been moved to a different file.
* Run experiments to predict real-valued tumor purity (regression), and analyze results in `02_classify_stratified/plot_purity_regression_results.ipynb`.


02_classify_stratified/nbconverted/plot_purity_regression_results.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Plot tumor purity prediction results
+# 
+# Here, we'll visualize the results of our tumor purity prediction. We're predicting real-valued purity labels here (e.g. elastic net regression).
```
miltondp, line 6: I know ElasticNet is pretty standard, but any reason why you chose this one in particular? Maybe a biological reason? How did you specify the regularization parameters?

02_classify_stratified/nbconverted/plot_purity_regression_results.py
```diff
@@ -0,0 +1,151 @@
...
+# 
+# Results analyzed here are generated by the `run_purity_prediction.py` script, without the `--classify` argument.
```
miltondp, line 8: So by default it runs regression, unless you specify `--classify`?

02_classify_stratified/nbconverted/plot_purity_regression_results.py
```diff
@@ -0,0 +1,151 @@
...
+import seaborn as sns
+
+import mpmp.config as cfg
+import mpmp.utilities.analysis_utilities as au
+
+
+# In[2]:
+
+
+# set results directory
```
miltondp, line 28: This is very minor, but usually, I structure my notebooks with section titles, such as "Setup" in this case, "Plot this and that", etc.

02_classify_stratified/nbconverted/plot_purity_regression_results.py
```diff
@@ -0,0 +1,151 @@
...
+axarr[1].set_title('Tumor purity prediction performance, by data type')
+axarr[1].set_xlabel('Signal or shuffled')
+axarr[1].set_ylabel(r'$R^2$')
+axarr[1].legend(title='Data type')
+
+
+# In[5]:
+
+
+# now facet by cancer type
```
miltondp, line 77: Very minor, same as before: I would add `# Facet by cancer type` here, so you can quickly move across the notebook results.

miltondp: I think you showing the top cancer types first, and then all the rest below. I would add section subtitles for those.

mpmp/utilities/analysis_utilities.py
```diff
@@ -100,7 +103,33 @@ def load_purity_binarized_results(results_dir):
     return results_df
 
 
-def load_purity_by_cancer_type(results_dir, sample_info_df):
+def load_purity_results(results_dir):
+    """Load results of tumor purity regression experiments.
```
miltondp, line 107: Does this function load regression results only?

mpmp/utilities/analysis_utilities.py
```diff
@@ -109,33 +138,38 @@ def load_purity_by_cancer_type(results_dir, sample_info_df):
...
         results_filename = str(results_file.stem)
-        if 'preds' not in results_filename: continue
+        if classify and ('classify' not in results_filename
+                         or 'preds' not in results_filename): continue
+        if not classify and ('regress' not in results_filename
+                             or 'preds' not in results_filename): continue
         if results_filename[0] == '.': continue
         if check_compressed_file(results_filename):
-            training_data = '_'.join(results_filename.split('_')[:-4])
+            training_data = '_'.join(results_filename.split('_')[:-5])
```
miltondp, line 158: (Minor) Not sure what's the template of the file name, but in the future, it might be better to use a regular expression here to avoid any issues with a wrong negative index.

mpmp/prediction/cross_validation.py
```diff
@@ -0,0 +1,279 @@
+"""
+Functions to split data for cross-validation.
```
miltondp, line 2: I will definitely take a look at these scripts in the future when I have to run any classificiation/regression task :)

mpmp/prediction/regression.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    # Get all performance results
+    y_predict_train = cv_pipeline.predict(X_train)
+    y_predict_test = cv_pipeline.predict(X_test)
+
+    return cv_pipeline, y_predict_train, y_predict_test, y_cv
+
+
+def get_preds(X_test_df, y_test_df, cv_pipeline, fold_no):
```
miltondp, line 88: This function seems very simple, but maybe a short description of what it does would be good.

mpmp/scripts/run_tumor_purity.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
miltondp, line 1: Maybe also here a short description of this script, that I think uses compressed and non-compressed data.

miltondp (approved):
Great, Jake, so it also works well by predicting continuous values. Not sure if you can compare AUPR with `R^2` or `1-RMSE` values, but is it accurate if we say that binarizing works better?

All the comments are really minor, and you can take them or not, your call.

As a general comment, maybe for a new project, if you think it's useful, an object-oriented design could be interesting for this kind of experiments. Not sure if in this case it would be helpful though.

jjc2718:
> Great, Jake, so it also works well by predicting continuous values. Not sure if you can compare AUPR with `R^2` or `1-RMSE` values, but is it accurate if we say that binarizing works better?

Good question - I do think it's hard to compare. My gut feeling would be that regression is a harder problem (RMSE values around 0.1 roughly means that on average the model misses the true value by about that much), whereas classification metrics for the best data types are approaching 1 (near perfect classification), but that's really a qualitative/high-level comparison. I'm not sure if there's a good way to quantify the difference between the binary and continuous problems.

> As a general comment, maybe for a new project, if you think it's useful, an object-oriented design could be interesting for this kind of experiments. Not sure if in this case it would be helpful though.

Yeah, I should really learn more about object-oriented ways to set up experiments like this. I think you do a good job of this in some of your projects - I'll have to look through some of your stuff, and maybe do some reading about this. Next time I start a new project it would be good to try!

02_classify_stratified/nbconverted/plot_purity_regression_results.py
jjc2718: Yep, exactly. I guess my thought there was that the natural way to think about predicting continuous purity values was using regression, so I decided to make that the default and add a flag to binarize the output.

mpmp/utilities/analysis_utilities.py
jjc2718: Yeah, good point - this filename parsing code is kind of a mess at the moment. Part of the problem is that some of the training datatype identifiers have underscores in them ("me_27k", "me_450k") and some don't ("expression"), and I'm also using underscores to separate parts of the filename.

Regex could be a solution, but I often find them harder to interpret and debug than just using basic Python string operations, so I try to do the latter whenever I can.

I'll create an issue to think about the filenames in the future.

02_classify_stratified/nbconverted/plot_purity_regression_results.py
jjc2718: Right, good question. A few observations about this:
* All of these datasets are high-dimensional (gene expression and methylation especially) and only a few of the features (genes or methylation probes) are likely to be relevant. Elastic net tends to be a good way to balance performance and feature selection for problems like this, and in similar past work it's tended to work well.
* If you're asking specifically about the sklearn `ElasticNet` vs. other ways of fitting a regularized regression model, I found it to be a good balance between runtime and accuracy. I also tried sklearn `SGDRegressor` and found it to be much more finicky/sensitive to parameter choices. `ElasticNet` uses coordinate descent rather than batch SGD to fit the model.
* For the parameters, I specify ranges to search over [here](https://github.com/greenelab/mpmp/blob/d2d805c82f15952d47f6c37bcfc704bc4d3c4258/mpmp/config.py#L85), then I'm doing a grid search on the training dataset to choose the best ones [here](https://github.com/greenelab/mpmp/blob/d2d805c82f15952d47f6c37bcfc704bc4d3c4258/mpmp/prediction/regression.py#L60). I chose the ranges based on [this sklearn documentation](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.ElasticNetCV.html), which suggests values close to 1 for `l1_ratio` and evenly spaced values for `alpha`. Other than that, I didn't do much tuning by hand.

I'll add a bit of documentation about the parameter ranges in `regression.py`, just for future reference.

mpmp/utilities/analysis_utilities.py
jjc2718: Currently, yes, but I think I can combine it with the function above (which does essentially the same thing for classification results), and just pass a classify flag to differentiate.

02_classify_stratified/nbconverted/plot_purity_regression_results.py
jjc2718: Good idea! I'll start doing this more consistently.

PR #37, opened 2021-03-30
Directory reorganization/cleanup

### LINKED ISSUE #25: Restructure experiment order to make more sense
See: https://github.com/greenelab/mpmp/pull/24#discussion_r582057696

jjc2718:
Closes #25.

This PR touches a lot of files but I'm just moving things around - there are very few functional changes here, and there shouldn't be much that needs in-depth review.

Major changes:
* Reorganize directory structure into discrete experiments (mutation prediction, cancer type prediction, tumor purity prediction)
* Move results from a single `results` directory into separate directories for each experiment, and update `config.py` and analysis scripts to take this into account
* Move bash scripts for running experiments to the corresponding experiment directory


miltondp (approved):
Looks good to me!

PR #38, opened 2021-04-05
Multi-omics mutation prediction

### LINKED ISSUE #3: Combining data modalities
See https://github.com/greenelab/mpmp/pull/2#discussion_r541016738

jjc2718:
**PR description:**

In these experiments, we want to compare elastic net logistic regression models using multiple data types to models using only a single data type. We're not doing anything particularly fancy here, just concatenating the feature sets (genes or CpG probes) from the individual data types to create a simple "multi-omics" model.

For now, we're just doing this for gene expression and the two methylation datasets, giving us 3 models with each of the pairwise combinations and 1 model with all three data types.

The main takeaway is that for this approach, the multi-omics models don't seem to perform much better than the single-omics models. For each target gene, this plot shows the best-performing single-omics vs. multi-omics model:

image

None of the differences are statistically significant.

**Major code changes:**

* Add function to `mpmp/utilities/data_utilities.py` to load multiple data types
* Add function to `mpmp/utilities/tcga_utilities.py` to preprocess multi-omics data sets (we're standardizing all of the gene/CpG features for these experiments, but keeping the cancer type and mutation burden covariates the same).
* Add scripts to `05_classify_mutations_multimodal` to generate and analyze results
* Add some simple unit tests in `tests/test_multimodal_preprocessing.py`

Closes #3.


05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -0,0 +1,304 @@
...
+# 
+# * In this last figure, we can see that the difference in distributions between the best-performing single-omics and multi-omics model is visually quite small for all of the target genes, and none of the differences are statistically significant using a t-test over the 8 total CV folds.
+# * Looking at the other two figures, when we consider all data types the difference between single-omics and multi-omics models is still pretty minimal in general. There are some genes where adding expression data helps a lot (EGFR, PIK3CA, KRAS) and others where it doesn't really (IDH1, SETD2), but in general there's not a lot of added predictive ability when expression is combined with other data types, suggesting that methylation may be most useful in that it correlates with expression patterns (although more follow-up would be needed to really confirm this).
+# * We had hypothesized that the reason for this might be that the multi-omics models are only selecting features from the best performing data type, but looking at the model coefficients this doesn't really seem to be the case; non-zero coefficients seem to be roughly evenly distributed between data types.
```
ben-heil, line 304: More data != more better :(

ben-heil (approved):
Looks good! I definitely feel your comments on result files getting messy. Mine all have like 5 different modifiers in their names now

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
nrosed, line 304: I'm not sure if having "non-zero coefficients seem to be roughly evenly distributed between data types" is evidence enough that each data type is being evenly used. It could be that the features LASSO is choosing are features that are highly correlated between the two data types. (unless you are doing some PCA/ICA on the whole dataset first and only using those features, then look at the loadings to see if the most informative dimensions are driven by methylation or expression).

What would be cool here would be if you could somehow breakdown the proportion of variance explained between the data types to really show how much more information each data type is adding. This is a bit tricky and I am not sure what measure is best, but I could imagine this being done by looking at model fit and seeing how it changes with adding more variables (maybe AIC? or likelihood ratio test?).

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -0,0 +1,304 @@
...
+sns.boxplot(data=plot_df, x='gene', y='delta_aupr', hue='model_type')
+plt.title('Best performing single-omics vs. multi-omics data type, per gene')
+plt.xlabel('Target gene')
+plt.ylabel('AUPR(signal) - AUPR(shuffled)')
+
+
+# Summary and interpretation:
+# 
+# * In this last figure, we can see that the difference in distributions between the best-performing single-omics and multi-omics model is visually quite small for all of the target genes, and none of the differences are statistically significant using a t-test over the 8 total CV folds.
+# * Looking at the other two figures, when we consider all data types the difference between single-omics and multi-omics models is still pretty minimal in general. There are some genes where adding expression data helps a lot (EGFR, PIK3CA, KRAS) and others where it doesn't really (IDH1, SETD2), but in general there's not a lot of added predictive ability when expression is combined with other data types, suggesting that methylation may be most useful in that it correlates with expression patterns (although more follow-up would be needed to really confirm this).
```
nrosed, line 303: is there any data-related differences between (EGFR, PIK3CA, KRAS) and (IDH1, SETD2)? I'm just curious if this could be driven by a confounder or if it's biologically driven. I'm mostly curious because I'm wondering if this is a dataset size issue; are EGFR, PIK3CA, KRAS more frequently mutated?

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -0,0 +1,304 @@
...
+    ax.set_ylabel('AUPR(signal) - AUPR(shuffled)')
+    ax.set_ylim(-0.2, max_aupr)
+    
+handles = []
+for ix, data in enumerate(data_order):
+    handle = mpatches.Patch(color=sns.color_palette()[ix], label=data)
+    handles.append(handle)
+    
+fig.delaxes(axarr[1, 3])
+plt.legend(handles=handles, loc='lower right')
```
nrosed, line 240: It's interesting that methyl27k does so much better than the other data types for TET2, but performs the worst for almost all other genes.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -0,0 +1,304 @@
...
+    ax.set_ylim(-0.1, max_aupr)
+    ax.legend_.remove()
+    
+handles = []
+for ix, data in enumerate(data_order):
+    handle = mpatches.Patch(color=sns.color_palette()[ix], label=data)
+    handles.append(handle)
+    
+fig.delaxes(axarr[1, 3])
+plt.legend(handles=handles, loc='lower right')
```
nrosed, line 91: just as a dumb sanity check, when you do the multi-modal models the performance is always increasing on the training set, right? I'm just surprised that the red boxplot is significantly below the blue for EGFR and PIK3CA. It would mean that the methylation data really has no information for those genes.

nrosed (approved):
Looks good to me! Only a few comments/questions added, but they are more about my curiosity, so feel free to ignore.

mpmp/utilities/tcga_utilities.py
```diff
@@ -272,6 +274,156 @@ def subset_by_mad(X_train_df, X_test_df, gene_features, subset_mad_genes, verbos
...
+                                data type, in the same order as indexing in
+                                data_types. Defaults to True for all data
+                                types.
+
+    Returns
+    -------
+    X_train_df (pd.DataFrame): preprocessed training data
+    X_test_df (pd.DataFrame): preprocessed test data
+
+    TODO write some tests, this logic is pretty complex
```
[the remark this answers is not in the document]
jjc2718: Good call, I forgot to remove this after writing the tests

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: Yeah, I feel like there's a lot of room for finding more sophisticated methods that would do better here (like the sort of approaches you're looking at for combining datasets), but it was a bit surprising that this didn't work. What we're doing is super naive, but we still expected the added data to help somewhat.

tests/test_multimodal_preprocessing.py
```diff
@@ -0,0 +1,88 @@
...
+import mpmp.config as cfg
+import mpmp.utilities.tcga_utilities as tu
+
+@pytest.fixture(scope='module')
+def raw_data():
+    """Generate a raw dataset"""
+    cols = list(string.ascii_lowercase)[:11]
+    X_train_raw_df = pd.DataFrame(np.random.uniform(size=(20, 11)), columns=cols)
+    X_test_raw_df = pd.DataFrame(np.random.uniform(size=(10, 11)), columns=cols)
+    data_types = np.array([0, 0, 0, 1, 1, 1, 2, 2, 2, -1, -1])
```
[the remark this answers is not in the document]
jjc2718: What I have now doesn't use a fixed index for each data type - it just assigns them in the order the data types are passed in. So if you wanted to build a model from expression data and 27K methylation data, it would assign those data types indexes 0 and 1 in that order.

I think I'll keep it the way it is since it's flexible (if I add other data types in the future it should generalize automatically), but I do like your idea of having a constant for non-gene features. I'll add that.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: This is a good point about the coefficients/feature correlations. What you said here:

> if you could somehow breakdown the proportion of variance explained between the data types to really show how much more information each data type is adding

is exactly what I'd like to do, but I agree that it seems tricky. We'll definitely think about ways to do it going forward (keeping your ideas in mind), although I'm not sure if it will make the cut for this paper.

Doing PCA or other dimension reduction on the whole dataset was also something we thought about, but we decided not to include it in this analysis/paper, at least for now. We really don't want to spend a ton of time optimizing this process at this point - our main goal for this analysis isn't necessarily to show that fitting a good multimodal model for this problem isn't _possible_, more just to show that it's not trivially easy.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: Yeah, I agree. I looked at TET2 because it's directly involved in DNA methylation (or more accurately DNA demethylation), but I'm hesitant to draw many conclusions from it because it's not that frequently mutated.

The cancer types that meet our cutoff for TET2 mutation (>10 samples mutated and >5% of samples mutated per cancer type) are  SKCM (31 samples, 8%) and UCEC (47 samples, 9%). SKCM has a very high baseline mutation burden so I'm not super convinced any of those samples are strongly driven by TET2 mutation (could be just chance/passenger mutations), and I can't find much work linking TET gene mutation to uterine/endometrial cancers (maybe [this](https://doi.org/10.1177/1010428317695017)?), although that association seems more plausible to me.

I still think it's interesting that performance is better with the methylation data here (since that's what we'd expect based on the function of TET2), but I'm interpreting with a bit of caution.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: Valid cancer types for each gene included in the train/test sets are here: https://docs.google.com/presentation/d/1pudohfOLGPa7DPIeyOIPZst-V0atzqaFdjKzuHC6neo/edit?usp=sharing

I guess the first three genes tend to be a bit more frequently mutated in some cancer types (particularly KRAS and PIK3CA), but IDH1 is very commonly mutated in LGG. Curious if you have any other thoughts based on those tables (`status_count` is the total number of mutated samples for that gene in that cancer type, and `status_proportion` is the proportion of mutated samples).

I did pick the first 3 genes and TP53 because we know they have a strong gene expression signature, and I picked IDH1 and SETD2 because they're involved in DNA methylation, so it's not exactly a random sample of genes. We could/might run this for more genes, but based on our previous results I'd expect the results to be similarly favorable to gene expression for other genes.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: This is a good question, thanks for pointing this out. Sorry in advance for the wall of text - I think this is important to think about.

Here's a plot of performance on the training set:

image

And the corresponding performance on the test set:

image

So my answer to your question would be "sometimes but not necessarily". In some cases performance does increase on the training set but not the test set (e.g. PIK3CA, SETD2), but in other cases (EGFR, IDH1) performance is lower on both the train and test sets for the multi-modal models.

I'm not sure whether or not to be concerned about this, though. The way that I'm choosing hyperparameters for the elastic net model is by nested cross-validation (e.g. run 3-fold cross-validation on the training set with each set of hyperparameters), then I choose the best set of hyperparameters based on mean validation set AUPR.

That is to say that if I was doing something like standard ordinary least squares I would expect training set performance to always increase as more features are added, but since I choose the model that performs the best on the validation set, it seems plausible that the best model could be one that underfits the training set slightly but performs well on held-out data, as opposed to one that (over)fits the training set more closely and performs slightly less well on held-out data.

But I think it's important to ask whether we're really finding a good model (and, like you said, the added data really isn't contributing anything), or if we're just underfitting to the multi-omics data or not choosing the right optimizer or hyperparameter ranges for the larger data dimension. I can think of a few ways to get at this:

* Check if for the _same set_ of hyperparameters, our training set error is lower for multi-omics data than for single-omics data (I think this should be true)
* Try running the same experiments with slightly larger hyperparameter ranges, or a different optimizer

If you have any other ideas, I'd definitely be interested in hearing them (if you made it this far). I created #41 to look more closely at this, and I think it's important to spend some time looking at this before we submit the paper; we really don't want to conclude that multi-modal data doesn't help if it's really just a model fitting/optimization issue.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
nrosed: Interesting! I think cautious interpretations are fine here.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
nrosed: > Check if for the same set of hyperparameters, our training set error is lower for multi-omics data than for single-omics data (I think this should be true)

i think this is a good approach, I also think this is true

> Try running the same experiments with slightly larger hyperparameter ranges, or a different optimizer

I think you can make a plot here to show loss vs hyperparamter value to see if you are finding the correct sweet-spot.  

I think it's an important aspect to make sure that you really identify or empirically show in some way that your model fit is optimal, I can just see it as a sticking point for a reviewer. I can't think of other ways to solidify it, but I'll let you know if I see something that could help.

jjc2718: Great, thanks, this is super helpful. I'm going to merge this PR, but I'll look at this more closely in another PR in the near future.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
nrosed: Yeah, I think this makes sense. I don't have further thoughts here, seeing that the cancer types are pretty diverse in each of the predictions and that it's not super biased in all mutations (except maybe IDH1), I think it looks good.

PR #42, opened 2021-04-19
Gene set comparison + random gene selection

### LINKED ISSUE #40: Match top mutated/random gene sets to size of Vogelstein dataset
For gene expression comparison experiments.

jjc2718:
**PR overview:**

For one of the figures in our paper, we want to compare mutation prediction in the set of cancer genes we've been using (from [Vogelstein et al. 2013](https://science.sciencemag.org/content/339/6127/1546)) with an equal number of highly mutated genes in TCGA, and an equal number of random genes. Before, we were only looking at the top 50 most mutated genes in TCGA, but we're extending this to the top 107 (to make the count equal to the Vogelstein gene set) and selecting 107 genes randomly. Closes #40.

The trend in the results is the same as before, with the Vogelstein gene set leading to substantially more significantly predictable genes than the other gene sets.

image

**Major changes:**

* Added a script to select random genes with 2 or more cancer types meeting our mutation count threshold (`00_download_data/nbconverted/sample_random_genes.py`)
* Updated `mpmp/utilities/data_utilities.py` to load larger gene sets
* Reran experiments to use new gene sets, and plot results in `02_classify_mutations/plot_expression_gene_sets.ipynb`


02_classify_mutations/nbconverted/plot_expression_gene_sets.py
```diff
@@ -0,0 +1,245 @@
...
+    
+    print('{}: {}/{} ({:.4f})'.format(
+        gene_set,
+        np.count_nonzero(data_results_df.reject_null),
+        data_results_df.shape[0],
+        np.count_nonzero(data_results_df.reject_null) / data_results_df.shape[0]
+    ))
+
+
+# In[8]:
```
ben-heil, line 205: Is there a reason that the gene distribution in the Vogelstein  is different from the distribution in the others? It feels as though the Vogelstein genes are all predictable, where some small subset of the others are.

data/random_genes.tsv
```diff
@@ -0,0 +1,108 @@
+gene	classification
+HECTD4	neither
```
ben-heil, line 2: To what extent are the genes whose mutations are predictable in random_genes/top_genes just the genes that are in the Vogelstein set?

ben-heil (approved):
Do you have an intuition for why the Vogelstein genes are easier to predict than the highly mutated group? Is it that the prior knowledge genes would have both many mutations and a large effect size?

02_classify_mutations/nbconverted/plot_expression_gene_sets.py
jjc2718: We were hoping to show that if we select genes based on prior knowledge that they're involved in cancer (the Vogelstein gene set), we can do better (i.e. shift the distribution upward toward higher AUPR values) compared to just selecting genes based on how frequently they're mutated.

So we think/hope that's the reason - just because a gene is frequently mutated in cancer doesn't mean it has any functional effect (they could be passenger mutations, e.g. because the gene is large). And, conversely, lots of strong drivers/highly predictable genes in the Vogelstein gene set aren't that frequently mutated, or are only mutated in a specific subset of samples.

data/random_genes.tsv
jjc2718: Yeah, good point - by eye it looks like a lot of them are, but I'll add some code to the notebook to count the overlap.

jjc2718: Here's the overlap between all genes in the gene sets:

image

And between the significantly predictable genes:

image

I haven't done any quantitative overlap analysis, but looking at the second diagram 12/43 significant genes in the most mutated gene set are also in Vogelstein, and 7/10 significant genes in the random gene set are also in Vogelstein. So for the random genes, most of the signal is also captured by the Vogelstein gene set, but maybe not so much for the most mutated genes (although some of it is).

jjc2718:
> Do you have an intuition for why the Vogelstein genes are easier to predict than the highly mutated group? Is it that the prior knowledge genes would have both many mutations and a large effect size?

Yeah, that would be my summary of what we're seeing. I mostly think it's driven by the effect size, though - a lot of the prior knowledge genes don't have _that_ many mutations (although the best predicted ones mostly do), but they should definitely have a larger effect size than many of the highly mutated genes.

Issue #44, opened 2021-04-22
where are the genes going


jjc2718:
As more data types are added, we end up with fewer genes (see https://github.com/greenelab/mpmp-manuscript/pull/8). I thought this was because of the inner CV loop ending up with folds with no positive samples, but it looks like sklearn GridSearchCV uses label-stratified CV by default (see https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html).

We should investigate why certain genes are non-deterministically (?) dropping out of certain data types, and think about if there's any way to preserve some of them across analyses.

jjc2718:
Examples of genes that are present in some, but not all, data types:

* GNAQ
* JAK3
* CDC73
* MPL

Looks like the variability is mainly in the shuffled files (e.g. some data type/seed combinations drop out for shuffled files, but signal files seem to be relatively consistent). Maybe we can fix the shuffling order for each random seed?

jjc2718:
It also looks like the cancer type filtering is based on the set of all samples, not the samples that are actually included in the experiment. We should probably be filtering gene/cancer type pairs independently for each sample set.

jjc2718:
Another inconsistency I noticed is that we're doing hyper-mutated sample filtering after we decide which diseases (and therefore genes) to include. This means that some genes may end up with considerably fewer than 15 mutated samples after filtering for hyper-mutators.

PR #48, opened 2021-04-27
Fix for "disappearing genes" issue

jjc2718:
This addresses an issue with certain genes that were present in the results for some seeds/data types, but not others. See #44 for some background and examples.

After exploring this, there ended up being a few separate issues, all of which are fixed in this PR:

* Shuffling the labels in the baseline models sometimes led to test sets that didn't have any mutated samples, especially for really imbalanced labels. Now I'm splitting train/test sets then shuffling them independently, to maintain the same label balance as in the un-shuffled models.
* I was using sets instead of lists in `mpmp/utilities/tcga_utilities.py` for the list of `valid_samples` - switched to passing around lists to make the order deterministic
* Also in `mpmp/utilities/tcga_utilities.py`, I was applying our cancer type mutation filter (>15 samples mutated, >5% samples mutated) before removing hyper-mutated samples. This led to some genes that had fewer than 15 positive labels in the final dataset, some of which would drop out further down the pipeline. I switched the order of the hyper-mutated samples filter and the cancer type filter, which makes things much more predictable/consistent.

The results of all these changes are that we end up with fewer valid genes total (85 for gene expression, 84 for methylation, 75 for all data types -> see `01_explore_data/missing_genes.ipynb`), but these genes are now consistent across data types and across signal/shuffled models within a given experiment. The end result is that our conclusions should be more robust and less prone to spurious signal/small sample sizes.

Looking at the genes that get dropped in the new experiments, most of them seem to be low signal/hard to predict (near the origin in our volcano plots):

image

I'm currently re-running all of our experiments with these updated gene sets, which will be a separate future PR (although I don't expect it to change our conclusions).


mpmp/prediction/cross_validation.py
```diff
@@ -89,6 +91,17 @@ def run_cv_stratified(data_model,
...
+        # shuffle labels for train/test sets separately
+        # this ensures that overall label balance isn't affected
+        # (see https://github.com/greenelab/mpmp/issues/44)
+        if shuffle_labels:
+            # we set a temp seed here to make sure this shuffling order
+            # is the same for each gene between data types, otherwise
+            # it might be slightly different depending on the global state
+            with temp_seed(data_model.seed):
+                y_train_df.status = np.random.permutation(y_train_df.status.values)
+                y_test_df.status = np.random.permutation(y_test_df.status.values)
```
ben-heil, line 103: Ooh this is really cool. I'll have to remember this logic for shuffling things in the future

01_explore_data/nbconverted/missing_genes.py
```diff
@@ -0,0 +1,268 @@
...
+    text_labels = []
+    a = pd.DataFrame({'x': x, 'y': y, 'gene': gene})
+    for i, point in a.iterrows():
+        if point['y'] > -np.log10(SIG_ALPHA / 1000):
+            text_labels.append(
+                ax.text(point['x'], point['y'], str(point['gene']))
+            )
+    return text_labels
+
+# plot mutation prediction from expression, in a volcano-like plot
```
ben-heil, line 244: Is there a reason the vanishing genes are all at the bottom of the distribution instead of uniformly distributed? Is it because genes with bigger effect sizes are better studied/show up in more data types?

ben-heil (approved):
Congrats on finding a set nondeterminism error, those are always a pain to track down

mpmp/prediction/cross_validation.py
jjc2718: Yeah, it's useful!

01_explore_data/nbconverted/missing_genes.py
jjc2718: The reason they were "vanishing" before is that there was only a few positive labels, so sometimes I'd (randomly) end up with a CV split with no positive labels, and AUROC/AUPR are undefined in that case. But I think because there were so few positive labels in these cases, we weren't able to train good classifiers anyway, so they were at the bottom because the label imbalance was so extreme.

So I do think what you're saying is the right idea, although I'd put it like this: the genes with large effect sizes in these experiments tend to have more positive labels/be more frequently mutated, so I'm not surprised that we're not losing any of them with this change.

It's an interesting thought, though - I'm sure there are some genes that are infrequently mutated and have a large effect size (can't think of an example off the top of my head), but I'm not sure our current classification approach would be effective at identifying those genes. Maybe an opportunity for better methods in the future!

PR #49, opened 2021-05-04
Figure generation script + other repo cleanup

jjc2718:
This PR might look like a lot of changes in terms of lines, but most of them are just cosmetic changes to figure generation and other small details.

Major changes:

* Added a script to generate publication figures (`generate_figures.ipynb`)
* Renamed scripts in `02_classify_mutations` to correspond more closely to which parts of the paper they're relevant to
* Changed how we're doing statistical testing slightly:
    * Before we were doing unpaired, equal-variance _t_-tests for all of our statistical comparisons (then correcting for multiple comparisons using FDR correction)
    * Now we're doing paired _t_-tests (since CV folds will have the same train/test labels, so fold 1, model 1 will be paired with fold 1, model 2, etc).

These aren't yet the final versions of the figures we'll use for the paper (I still have to finish running some things and update the figures), but this PR will put in place the infrastructure for regenerating the publication figures once the data is generated.


mpmp/utilities/analysis_utilities.py
```diff
@@ -433,48 +453,62 @@ def compare_control_ind(results_df,
...
-                      (single_cancer_df.signal == 'signal'))
-        single_cancer_results = single_cancer_df[conditions][metric].values
-
-        conditions = ((pancancer_df[identifier] == id_str) &
-                      (pancancer_df.data_type == 'test') &
-                      (pancancer_df.signal == 'signal'))
-        pancancer_results = pancancer_df[conditions][metric].values
-
-        if single_cancer_results.shape != pancancer_results.shape:
+        conditions = ((condition_1_df[identifier] == id_str) &
```
ajlee21, line 469: For your statistical comparison when you compare model 1 (expression model) fold 1 with model 2 (27K methylation model) fold 1, does that mean that you have the same samples in fold 1? So you have the same seed you're using here?

By condition here do you mean model?

ajlee21 (approved):
Nice nice figures. @danich1 is a trend setter!

Some comments on the figures as you finalize them, so these changes don't need to be made in this PR:

* Panels and text for title and axis are fairly small
* (super nitpicky) Not sure if you want to change the colors for your "volcano" plots since the same colors are used in the box plots but they aren't associated with each other
* Looks like you change colors for your data type in figure 6

For your statistical comparison when you compare model 1 (expression model) fold 1 with model 2 (27K methylation model) fold 1, does that mean that you have the same samples in fold 1? So you have the same seed you're using here?

jjc2718:
> Nice nice figures. @danich1 is a trend setter!

Yeah, `svgutils` is so much easier than having to remake Inkscape figures every time I change a panel! Super cool stuff.
 
> Some comments on the figures as you finalize them, so these changes don't need to be made in this PR:
> 
> * Panels and text for title and axis are fairly small
> * (super nitpicky) Not sure if you want to change the colors for your "volcano" plots since the same colors are used in the box plots but they aren't associated with each other
> * Looks like you change colors for your data type in figure 6

These are all great points. The figures are definitely not finalized yet (I still have to tweak/rerun some things) but I'll keep these in mind as I remake them.

Created #50 to remind myself.

mpmp/utilities/analysis_utilities.py
jjc2718: > For your statistical comparison when you compare model 1 (expression model) fold 1 with model 2 (27K methylation model) fold 1, does that mean that you have the same samples in fold 1? So you have the same seed you're using here?

Yep, the folds have the same train/test samples across data types. This was something I made sure of as part of #48.

> By condition here do you mean model?

For all of my experiments I do mean model (i.e. comparing a model trained on data type 1 with a model trained on data type 2). I wanted to write this function in a general way because I can imagine using it for other comparisons in the future (for example, we could use it to compare single-cancer and pan-cancer models, like we were doing in https://github.com/greenelab/pancancer-evaluation), which is why I used "condition" instead of "model" - I feel like it's a more general term.

PR #51, opened 2021-05-17
Refactor plotting code in notebooks

### LINKED ISSUE #18: Refactor code in compressed results notebook
See: https://github.com/greenelab/mpmp/pull/17#discussion_r569892925

jjc2718:
Closes #18.

There's currently a lot of duplicated code in the notebooks I'm using to plot my results. I'm making similar types of plots between experiments, so I decided to refactor the code to make these common plots into shared functions.

This cuts down considerably on the complexity of the plot notebooks and creates a single point of truth for some of the details (so if I change these plots in the future I only have to do it in one place, rather than in each notebook individually).


mpmp/utilities/analysis_utilities.py
```diff
@@ -562,3 +562,36 @@ def process_coefs(coefs_df):
     return id_coefs
 
 
+def compare_all_data_types(results_df,
+                           sig_alpha,
+                           filter_genes=True,
+                           identifier='identifier',
+                           metric='aupr'):
+    """Run compare_results for each data type + assemble into dataframe"""
```
ajlee21, line 570: I know you have several difference comparisons (i.e. your volcano plot and then your bar plots), maybe add a bit more detail here on what the plot is - "Compare the ability of each data type (i.e. gene expression, methylation) to predict if a gene is mutated or not. Return dataframe with AUPRC values per seed."

mpmp/utilities/plot_utilities.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def plot_volcano_comparison(results_df,
+                            axarr,
+                            training_data_map,
+                            sig_alpha,
+                            xlim=None,
+                            ylim=None,
+                            verbose=False):
+    """Make a scatter plot comparing classifier results to expression.
```
ajlee21, line 105: Will this always compare against expression data? Or are you planning to vary the reference dataset you use?

mpmp/utilities/plot_utilities.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Arguments
+    ---------
+    results_df (pd.DataFrame): dataframe with processed results/p-values
+    axarr (array of Axes): axes array from plt.subplots
+    training_data_map (dict): maps internal data type names to display names
+
+    Returns
+    -------
+    axarr (array of Axes): axes with plots
```
ajlee21, line 219: I'm a little lost here. So is this returning a multi-panel figure? In this case it looks like there are 2 panels..?

ajlee21 (approved):
Very nice, I'm sure this will greatly help you to debug your code in the future and make the code more clean and usable as well.

Just had a couple of minor comments

mpmp/utilities/plot_utilities.py
jjc2718: For now, we're always comparing against expression data, since doing all the comparisons would be lots of plots (in the case of all the data types it would be 6 choose 2 = 15 plots).

In the future this wouldn't be hard to change if necessary, but for now I'll keep it specifically to expression.

mpmp/utilities/analysis_utilities.py
jjc2718: Good idea, I got a bit lazy with this documentation. I'll add more detail.

mpmp/utilities/plot_utilities.py
jjc2718: Yeah, exactly, this returns an array of the matplotlib axes in the multi-panel figure.

I actually don't think this function needs to return anything, because Python objects (like matplotlib `Array` or `Figure` objects) are mutable. This means if we make changes to them inside the function (like adding plots to the axes), the changes will persist outside of the function.

This is as opposed to something like a Python string, which is immutable so you'd have to reassign it like I'm doing here.

jjc2718:
Tests are failing because of an ongoing issue with GitHub Actions (see: https://www.githubstatus.com/)

I'll wait a bit to see if the issue resolves, but if not I'll just merge. I ran the tests locally and they all passed.

PR #52, opened 2021-05-20
Plot heatmap to summarize per-gene comparison results

jjc2718:
We think it would be helpful to have a figure summarizing which data type performs best for each of the genes that we're using for prediction. I created these figures both for the expression vs. DNA methylation comparison and for the comparison including all data types, but we'll most likely use the all data types figure for the paper.

image

Blue highlight = classifier trained on that data type outperforms the baseline with permuted labels, red highlight = data type outperforms baseline and outperforms other data types for that gene


ben-heil (approved):
Looks good! One small aesthetic note, it may be easier to parse the plot if you sort the datatypes by number of red or blue boxes. If you're using the current order for the other plots it probably wouldn't be worth the inconsistency though

jjc2718:
> Looks good! One small aesthetic note, it may be easier to parse the plot if you sort the datatypes by number of red or blue boxes. If you're using the current order for the other plots it probably wouldn't be worth the inconsistency though

Good point - currently I'm sorting by lexical order (which is the Seaborn heatmap default) but I think I want the data types to be in the same order as the other plots. I'll try that and sorting by number of boxes, and see which one looks better.

PR #53, opened 2021-06-01
Update figures with results from extended parameter ranges

### LINKED ISSUE #44: where are the genes going

jjc2718:
Closes #44 (finally - these results should be pretty close to what we want to submit at this point).

The main change in this PR is rerunning all of the single-omics experiments with the correct gene sets and larger hyperparameter ranges (see changes to `mpmp/config.py`), and updating all of the figures with the new results.

The change in hyperparameter ranges caused a fairly substantial change to our results and interpretation: the methylation datasets now perform a lot better, generally on par with gene expression data. We think this is probably because the parameters were optimized at some point in Greg's papers to work well with expression data, so expanding the ranges favored the other data types.

Most of the code changes are related to the figures (either refactoring the plotting code to be a bit simpler, or adding information/details to the figures). Nothing too consequential other than the parameter range updates.


02_classify_mutations/scripts/run_methylation.sh
```diff
@@ -1,20 +1,36 @@
...
 N_DIM=5000
 
 mkdir -p $ERRORS_DIR
 
-for data_type in expression me_27k_bmiq; do
-    cmd="python 02_classify_stratified/run_mutation_classification.py --gene_set vogelstein --results_dir $RESULTS_DIR --training_data ${data_type} --seed $SEED 2>$ERRORS_DIR/errors_${data_type}.txt"
-    echo "Running: $cmd"
-    eval $cmd
-done
+for seed in 42 1; do
```
ben-heil, line 8: I thought this line was python for a second, and it caused my brain to melt

ben-heil (approved):
Looks good! Figures 1 and 6 aren't displayed in `generate_figures.ipynb`, though that may be intentional.

jjc2718:
> Looks good! Figures 1 and 6 aren't displayed in `generate_figures.ipynb`, though that may be intentional.

Yep, that's intentional (Figure 1 is generated in BioRender/Inkscape since it's an overview figure/schematic thing, and Figure 6 is just directly generated by another notebook with no panel wrangling required). I'll make this clear in the notebook.

PR #56, opened 2021-06-22
Figure edits

### LINKED ISSUE #50: Tweaks to figures
See: https://github.com/greenelab/mpmp/pull/49#pullrequestreview-651592845

jjc2718:
Closes #50.

This PR makes some small changes to the figures and how they're generated, both from a previous PR (see linked issue) and from some collaborator feedback (see greenelab/mpmp-manuscript#24).

The major changes to the code are:
* Updating the multi-omics plot script to generate results for raw and compressed features (see `05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py`)
* Updating the heatmap script to plot all statistically equivalent predictors as the "best" predictor, rather than just the one with the highest mean performance (see changes to `mpmp/utilities/plot_utilities.py`)

Additional changes are mostly small tweaks to color schemes, etc.


05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -105,104 +112,67 @@
...
-print(u_results_df_2.training_data.unique())
-print(u_results_df_3.training_data.unique())
-print(u_results_df_4.training_data.unique())
-print(u_results_df_1.seed.unique())
-print(u_results_df_2.seed.unique())
-print(u_results_df_3.seed.unique())
-print(u_results_df_4.seed.unique())
+    # make sure data loaded matches our expectations
+    print(u_results_df.training_data.unique())
+    print(u_results_df.seed.unique())
```
ben-heil, line 129: You may want to make these assert statements instead of print statements if their goal is to make sure nothing is changing behind the scenes

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
```diff
@@ -27,19 +27,26 @@
...
-    'results',
-    'pilot_results_all_feats',
-    'gene'
-).resolve()
-
 # if True, save figures to ./images directory
 SAVE_FIGS = True
 
+# if True, use raw feature results, else use compressed features
+RAW_FEATS = True
```
ben-heil, line 34: Is the idea that there are two datasets (or feature spaces I  guess) and users will have to run the notebook twice with `RAW_FEATS` set to True and then False to see both sets of figures? If so, why not just do two notebooks, or a function with a flag that you call twice so that all the figures are generated without manual intervention?

mpmp/utilities/plot_utilities.py
```diff
@@ -264,34 +281,158 @@ def plot_heatmap(heatmap_df, results_df):
     cbar.outline.set_edgecolor('black')
     cbar.outline.set_linewidth(1)
 
+    ax = plt.gca()
```
ben-heil, line 284: I love how matplotlib (or I guess seaborn in this case) code reads like C code. Who needs full word variable names anyway hahaha

ben-heil (commented):
Looks good! If the RAW_FEATS switch is there to allow future users to use other features than what you are currently working with then treat this review as approving. If they need to make manual changes to the notebook flags in order to get all the figures to generate, then I disagree with that design decision and we should probably talk about it

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: The figures generated with `RAW_FEATS=False` are in the main paper, and the figures generated with `RAW_FEATS=True` go in the supplement (which I guess should probably be in a comment somewhere). Other than the results data the notebook pulls from, there's no difference in how the figures are generated.

I guess I could make this into two notebooks, but that would result in a lot of duplicate code that I'd have to update in two places anytime I change things. Maybe I can find a way to call a function twice with a flag and generate all the figures at once - let me think about it.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
jjc2718: These are less to make sure _nothing_ is changing, and more just to visualize what's changing as I apply this script to different sets of results. The print statements allow me to see at a glance how many replicates/CV folds I ran (and make sure it's what I expected), as well as what models I ran (and make sure they're what I expected as well).

I guess since these results are more or less finalized for the paper now, I'm open to taking these out - let me know what you think.

jjc2718:
I refactored the plot script `05_classify_mutations_multimodal/plot_multimodal_results.ipynb` to plot the results with both the compressed features (main paper) and raw features (supplementary figures), so they should both happen anytime the script runs. Let me know if this seems cleaner.

05_classify_mutations_multimodal/nbconverted/plot_multimodal_results.py
ben-heil: I think leaving them in or removing them is fine. The comment just gave me the impression that the statements were supposed to always return the same thing.

ben-heil (approved):
The new notebook format looks much cleaner!

PR #58, opened 2021-07-09
Small changes to plots/statistical testing code

jjc2718:
These changes correspond to the paper updates in greenelab/mpmp-manuscript#26 .

The main changes here are:
* Add additional dotted lines for other significance thresholds to volcano plots
* Add pairwise statistical tests to box plots (we're using a paired Wilcoxon test for these, with a Bonferroni multiple testing correction).
* Also use a paired Wilcoxon test to compare the multi-omics models (results don't change at all, just for consistency)

None of the results or interpretations are changing here in any substantial way, the statistical tests are just intended to make our conclusions a bit more quantitative (rather than saying "gene expression looks a little bit better", etc).


05_classify_mutations_multimodal/plot_multimodal_results.ipynb
```diff
@@ -845,12 +848,12 @@
...
-      "KRAS diff: -0.0013 (pval: 0.9424)\n",
-      "PIK3CA diff: -0.0054 (pval: 0.8047)\n",
-      "SETD2 diff: -0.0070 (pval: 0.8165)\n",
-      "TP53 diff: -0.0037 (pval: 0.7065)\n"
+      "EGFR diff: -0.0286 (pval: 0.1484)\n",
+      "IDH1 diff: 0.0040 (pval: 0.8438)\n",
+      "KRAS diff: -0.0013 (pval: 0.8438)\n",
+      "PIK3CA diff: -0.0054 (pval: 0.6406)\n",
+      "SETD2 diff: -0.0070 (pval: 0.6406)\n",
+      "TP53 diff: -0.0037 (pval: 0.6406)\n"
```
ben-heil, line 856: I probably don't understand because I'm a garbage statistician, but is there a reason that you see the same p-values across tests? Is it because it's a ranked test?

ben-heil (approved):
I like the alpha value bars in the AUPR plots!

05_classify_mutations_multimodal/plot_multimodal_results.ipynb
jjc2718: It is because it's a ranked test - that means there's only a discrete number of possibilities for test statistics/p-values, and since there's only n=8 for these tests the odds of seeing the same p-value for 2 comparisons are fairly high.

Just to verify, I wrote a quick Python script to calculate the test statistics by hand for the duplicated p-values here, using [the description on Wikipedia](https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test#Test_procedure):

Screen Shot 2021-07-12 at 1 25 21 PM

(the distribution of the test statistic is symmetric so +/- 4 result in the same p-value). So the repeated p-values do have the same test statistic, since the rank * sign combinations end up being equivalent.

PR #59, opened 2021-07-28
Comparing -omics types for survival prediction, part 1

jjc2718:
**PR description:**

As we've been working toward preprinting/submitting our paper, one idea a collaborator brought up was that it could strengthen our arguments to look at more than just mutation prediction. A common problem of interest in cancer genomics/multi-omics is predicting survival duration (sometimes referred to as prognosis prediction).

This is a bit more complicated than the standard classification and regression problems that we've been looking at due to the presence of "censored" data. In other words, when a patient has a death recorded you know how to label the sample, but when there's no death recorded, there's often no way to know if the patient is still alive, or if they just withdrew from the study or lost touch with their doctors. There's a nice explanation of challenges and modeling strategies [in the scikit-survival documentation](https://scikit-survival.readthedocs.io/en/stable/user_guide/00-introduction.html), which is the Python package we used.

In general, we used the same basic approach as we used for mutation prediction (comparing individual -omics types), but using [elastic net Cox regression](https://scikit-survival.readthedocs.io/en/stable/user_guide/coxnet.html) instead of logistic regression.

**Results:**

So far, we've only run this with the top 1000 features for the expression and methylation data types. We're planning to run this for all the data types using both raw and PCA compressed features; those results will come in the next PR, hopefully early next week.

In general, gene expression seems to outperform the methylation data types slightly, but predictive ability is fairly comparable between data types. (Red square = significantly outperforms permuted-labels baseline + statistically equivalent to best-performing data type)

image

**Main code changes:**

* `06_predict_survival/run_survival_prediction.py`: script to run experiments
* `06_predict_survival/plot_survival_results.ipynb`: script to plot results
* `mpmp/prediction/survival.py`: main modeling/prediction code
* `load_survival_labels` in `mpmp/utilities/data_utilities.py`: main data loading/preprocessing code


06_predict_survival/nbconverted/plot_survival_results.py
```diff
@@ -0,0 +1,207 @@
...
+
+sns.set({'figure.figsize': (28, 6)})
+sns.set_context('notebook', font_scale=1.5)
+
+ax = plu.plot_heatmap(heatmap_df,
+                      all_results_df.reset_index(drop=True),
+                      different_from_best=True,
+                      raw_results_df=raw_results_df,
+                      metric='cindex',
+                      id_name='cancer_type')
```
ben-heil, line 204: Transcriptomics = best omics

I know previously you found that a model trained on a concatentation of 'omics performed <= just expression. Would you expect the same to be true here? Do you have an intuition for why expression and methylation seem to be conveying the same information?

06_predict_survival/run_survival_prediction.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # argument group for parameters related to input/output
+    # (e.g. filenames, logging/verbosity options, target genes)
+    #
+    # these don't affect the model output, and thus don't need to be saved
+    # with the results of the experiment
+    io = parser.add_argument_group('io',
+                                   'arguments related to script input/output, '
+                                   'note these will *not* be saved in metadata ')
+    io.add_argument('--cancer_types', nargs='*', default=None,
```
ben-heil, line 43: It's a little bit unintuitive that None = all. I guess it makes sense to do this way because there might be a string collision between a cancer type called 'ALL' and the desired function of running all cancer types

mpmp/utilities/data_utilities.py
```diff
@@ -331,6 +331,84 @@ def load_purity(mut_burden_df,
...
+    Returns
+    -------
+    survival_df (pd.DataFrame): dataframe where the "status" attribute is
+                                the relevant survival duration
+    """
+
+    if verbose:
+        print('Loading survival info and covariates...', file=sys.stderr)
+
+    # TODO: drop relevant NA columns later
```
ben-heil, line 355: Is this TODO something you wanted to do before the paper goes out, or just something to look at eventually?

ben-heil (approved):
Looks good! I assume you don't have standard clinical data for the patients like age/sex that could be used to predict survival to see if 'omics data goes beyond that?

nrosed:
Looks good! Few small comments:

1. Have you tried using cancer subtype/stage as a confounder? Survival is highly affected by stage at which the patient was diagnosed.
2. Not sure if you already inspected these, but for some of the models (high and low performing) it's sometimes useful to plot the survival curves, then color by expression of the top markers + other confounders? Just to make sure it's working and to get a sense of the variation across patients. It's a bit tricky when you have so many variables in the survival model, but it may also help to plot the expression of the top markers comparing between the patients in the longest and shortest surviving groups.
3. If you are having sensitivity to hyperparameters, you could try running your model on a very homogeneous subset of data, luminalA/luminalB BRCA is both large and more or less uniform. My thinking is that in this subset of samples you won't get small subgroups driven by a noisy set of genes or driven by confounders.

nrosed (approved):

06_predict_survival/run_survival_prediction.py
jjc2718: Yeah, I guess 'ALL' would be a bit ambiguous (although there isn't any ALL data in TCGA unfortunately) but another string like `'all_cancer_types'` might be clearer.

mpmp/utilities/data_utilities.py
jjc2718: Neither, this was a TODO I added during development but forgot to take out when I addressed it. The code on lines 396-398 should handle all the NAs, so I'll remove the TODO.

06_predict_survival/nbconverted/plot_survival_results.py
jjc2718: Personally, I would expect it to be the same here, although I don't think we're going to try the multi-omics models for survival prediction right now (we want to get this paper submitted soon and it already has a ton of data/figures in it).

I think the reasons could be either biological (driver mutations with strong roles in cancer development lead to everything getting messed up, resulting in signatures of cancer dysregulation in both DNA methylation/epigenetics and gene expression patterns; also promoter methylation affects gene expression so we'd expect to see a bit of a causal link) or technical (we're using a really simple multi-omics method of just concatenating datasets, and it's possible that more sophisticated methods would align the signal better/extract the unique information more effectively in each data layer).

Distinguishing between the two is definitely a topic we're interested in for future work, but probably beyond the scope of this paper.

jjc2718:
Response to @nrosed comments:

> 1. Have you tried using cancer subtype/stage as a confounder? Survival is highly affected by stage at which the patient was diagnosed.

Not yet - our collaborator suggested including age as a covariate (along with our usual mutation burden and cancer type in the pan-cancer analysis), but including stage at diagnosis would also be a good idea.

> 2. Not sure if you already inspected these, but for some of the models (high and low performing) it's sometimes useful to plot the survival curves, then color by expression of the top markers + other confounders? Just to make sure it's working and to get a sense of the variation across patients. It's a bit tricky when you have so many variables in the survival model, but it may also help to plot the expression of the top markers comparing between the patients in the longest and shortest surviving groups.

This is a good idea! I'll try to plot a few survival curves with relevant covariates/markers for my next PR, as a sanity check. Looking at coefficients, we definitely see that age has a strong negative correlation with survival in just about every cancer type, so that should be easy to visualize in a plot.

> 3. If you are having sensitivity to hyperparameters, you could try running your model on a very homogeneous subset of data, luminalA/luminalB BRCA is both large and more or less uniform. My thinking is that in this subset of samples you won't get small subgroups driven by a noisy set of genes or driven by confounders.

Good to know! I'll give this a try if we continue to see convergence/sensitivity issues.

jjc2718:
Response to @ben-heil general comment:

> Looks good! I assume you don't have standard clinical data for the patients like age/sex that could be used to predict survival to see if 'omics data goes beyond that?

We do have some clinical data, and we're including age as a covariate on the recommendation of a collaborator. We haven't been including sex, but that may also be a good idea to try.

We did think about comparing to a clinical features-only baseline (which has proven surprisingly hard to beat in other survival prediction work, e.g. [this paper](https://doi.org/10.1093/bib/bbaa167)), but the focus of this work isn't so much to develop the *best* survival predictor that we possibly can, more just to compare the content of different -omics types against one another for this problem. Lots of other studies have benchmarked different -omics integration and clinical feature extraction methods for survival prediction, so we don't feel the need to get into that kind of comparison here.

PR #60, opened 2021-08-05
Microsatellite instability (MSI) prediction

jjc2718:
As a positive control for the mutational signatures data, we wanted to try predicting [microsatellite instability](https://en.wikipedia.org/wiki/Microsatellite_instability) status of TCGA samples from the -omics datasets that we're using for mutation prediction. MSI classes were experimentally determined in a few cancer types where it occurs frequently (COAD/READ, STAD, UCEC) from targeted sequencing panels as part of the TCGA Pan-Cancer Atlas; in other cancer types it's very rare and probably wouldn't be worth training a classifier for.

In [previous work](https://doi.org/10.1371/journal.pcbi.1008878), this has been a fairly straightforward classification problem where genomic data generally performs well. In addition, [the COSMIC single-base signatures](https://cancer.sanger.ac.uk/signatures/sbs/) that we're using have a few annotations for microsatellite instability, so we wanted to see if those show up as strong predictors.

Our results are pretty much what we expected, but MSI classification performance does show that standardizing the mutational signatures data before classification seems to help, so we're planning to make that change for the mutation prediction experiments as well.

image


jjc2718:
Looks like GitHub Actions is down right now, which is why the tests are failing - I'll rerun them when it's back up. https://www.githubstatus.com/incidents/7p1nnvkgh96y

04_predict_controls/nbconverted/plot_msi_results.py
```diff
@@ -0,0 +1,179 @@
...
+
+sns.boxplot(data=plot_df, x='identifier', y='auroc', hue='training_data', ax=axarr[1])
+axarr[1].set_title('MSI prediction performance, AUROC')
+axarr[1].set_xlabel('Cancer type')
+axarr[1].set_ylabel('AUROC')
+axarr[1].set_ylim(0.35, 1.05)
+axarr[1].legend(title='Data type')
+
+
+# ### Plot results per cancer type, corrected for permuted labels baseline
```
ajlee21, line 77: What do you mean here that you corrected for permuted labels baseline?

04_predict_controls/nbconverted/plot_msi_results.py
```diff
@@ -0,0 +1,179 @@
...
+            nz_coefs.append([identifier, seed, fold_no, feat_name, feat_coef])
+            
+nz_coefs_df = pd.DataFrame(nz_coefs, columns=['identifier', 'seed', 'fold', 'name', 'coef'])
+nz_coefs_df.head()
+
+
+# In[8]:
+
+
+# just average each coefficient over seeds/folds for now
```
ajlee21, line 150: Can you compare the coeff/weight from your regression model across your different cancer models? I guess you can at least say if they are in the upper quantile of the distribution of weights for your model..

ajlee21 (approved):
This looks really interesting, one basic cancer question: Why is predicting MSI a positive control for predicting mutations?

I left some other comments and questions below

04_predict_controls/nbconverted/plot_msi_results.py
jjc2718: Yeah, I think this would be interesting to dive deeper into. I think for this paper we're not interested in going too far into how the feature selection works (other than the basic sanity checks in this NB), but after this paper is submitted I'm tentatively planning to look further into some of these things as a start to my next project. Stay tuned...

04_predict_controls/nbconverted/plot_msi_results.py
jjc2718: In most of our mutation prediction experiments, instead of directly reporting classification metrics (AUPR, AUROC etc), we report the difference in these metrics vs. a model where the labels are permuted. The idea behind this is that the permuted-labels models are a simple way to correct for differences in the baseline frequency of different mutations (or in this case, in the baseline frequency of MSI status across different cancer types).

For example, in cases where a gene is very frequently mutated, the permuted labels baseline will do well just by chance (i.e. just by predicting "mutation" every time), so just looking at raw classification metrics may give an inflated score for the true predictor in those cases. If the gene is mutated around 50% of the time (balanced labels) this is less of an issue, but in many cases that's not true for us, unfortunately.

I'll add a note here about this, so the notebook will make more sense on its own (we have a discussion of this in the methods section of our paper, and I think I made notes in some of the mutation prediction notebooks too).

jjc2718:
> Why is predicting MSI a positive control for predicting mutations?

Sorry, maybe this was a little confusing because we've been talking about predicting mutations for so long. We're using MSI prediction as a control for the *mutational signatures* -omics features (one of the data types we used in addition to expression, DNA methylation, etc). Using mutational signatures performed really poorly for the mutation prediction task, so we wanted to make sure that we could use them for a different prediction task where we know they should work well.

Tumors that are microsatellite instability-high (our positive class in these experiments) typically have mutations in DNA damage repair genes, which lead to distinct hypermutation patterns in "microsatellite" DNA fragments (and elsewhere, but the microsatellites are easy to find and assay using PCR so they're a convenient clinical marker of defective DNA repair). Since these patterns affect DNA directly, mutational signatures (essentially a reduced-dimensional version of the tumor's somatic mutation profile) should be a more direct readout than things like gene expression and DNA methylation.

So that's a very long-winded of saying that MSI prediction is a different problem (i.e. different set of labels) than mutation prediction, but should be a good control for the mutational signatures *input data*. Hope that clarifies a bit!

04_predict_controls/nbconverted/plot_msi_results.py
ajlee21: That makes sense thanks for clarifying

PR #62, opened 2021-09-03
Comparing -omics types for survival prediction, part 2

jjc2718:
Problem description and previous changes in #59. Sorry this is a fairly large PR, no rush to get it reviewed so take your time.

**Main code changes:**

* I changed how we're extracting PCA features a bit for various data types: main changes are in `mpmp/utilities/data_utilities.py` and `mpmp/utilities/tcga_utilities.py`. Basically, this checks if PCA features have already been generated for a given data type, and if not we calculate and save them to a .tsv file.
* Code to generate survival curves in `mpmp/prediction/cross_validation.py` and `mpmp/prediction/survival.py`.
* Notebooks to visualize survival curves (`06_predict_survival/plot_survival_curves.ipynb`) and to compare PCA dimension (`06_predict_survival/plot_survival_pc_comparison.ipynb`)
* Updated data download scripts in `00_download_data` to use the same PCA compressed features as the classification scripts; these probably don't need to be reviewed extensively

**PR description:**

This should wrap up the changes for survival prediction, which we are planning on including in our paper (manuscript edits to come). Across data types, we see better model stability and more reasonable results when we use the top principal components as features, and we decided to use PCA-derived features for all data types to make per-feature information content comparable.

**Results:**

We decided to stick with PCA compressed features for the paper, and we show results for several different numbers of principal components, in addition to age and mutation burden as covariates in all the models. In general, we see that expression and methylation are fairly comparable for pan-cancer survival prediction (similar to our mutation prediction results). We also see that RPPA data performs fairly well, with miRNA and mutational signatures performing relatively poorly.

image

We also looked at some survival curves, as suggested by @nrosed in #59, in the `plot_survival_curves.ipynb` notebook. For some cancer types (BRCA, LGG) predicted survival trended closely with age and subtype, and for others (LUAD) there wasn't much of a trend. You can see a few examples in the first few slides of this presentation: https://docs.google.com/presentation/d/1Kh6QUTadjk9RSQDXIIvfHF50a9cZb8uYmmqgwfeaUIg/edit?usp=sharing


06_predict_survival/scripts/run_survival.sh
```diff
@@ -1,23 +1,37 @@
...
-        cmd+="--training_data ${data_type} "
-        cmd+="2>$ERRORS_DIR/errors_${data_type}.txt"
-        echo "Running: $cmd"
-        eval $cmd
-    done
+        # raw data
+        results_dir=./06_predict_survival/results/results_${n_dim}_top_mad/
+        for data_type in expression me_27k me_450k; do
+            cmd="python 06_predict_survival/run_survival_prediction.py "
+            cmd+="--results_dir $results_dir "
```
ben-heil, line 12: Oh cool, I didn't realize there was a concatenation operator in Bash. I thought you  had to do `cmd="${cmd}--results_dir $results_dir"`

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
```diff
@@ -0,0 +1,459 @@
...
+
+sns.set({'figure.figsize': (8, 6)})
+sns.set_style('whitegrid')
+
+sns.pointplot(data=me_performance_df, x='n_dim', y='delta_cindex', hue='training_data',
+              hue_order=training_data_map.values())
+plt.xlabel('Number of PCs')
+plt.ylabel('cindex(signal) - cindex(shuffled)')
+plt.title('Performance for varying PC count, averaged over cancer types')
+
```
ben-heil, line 177: Transcriptome = best 'ome

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
```diff
@@ -0,0 +1,459 @@
...
+sns.set_style('whitegrid')
+
+sns.pointplot(data=me_performance_df[me_performance_df.cancer_type == 'pancancer'],
+              x='n_dim', y='delta_cindex', hue='training_data', 
+              hue_order=training_data_map.values())
+plt.xlabel('Number of PCs', size=14)
+plt.ylabel('cindex(signal) - cindex(shuffled)', size=14)
+plt.legend(title='Training data', fontsize=13, title_fontsize=13)
+plt.title('Pan-cancer survival performance, expression/methylation', size=14)
+plt.ylim(-0.05, 0.25)
```
ben-heil, line 211: Just making sure I'm reading this right, pan-cancer is training models using all cancer types, while the previous figure trains models on each individual type then averages across cancers

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
```diff
@@ -0,0 +1,459 @@
...
+              hue='training_data', hue_order=training_data_map.values(), ax=axarr[1])
+axarr[1].set_xlabel('Number of PCs')
+axarr[1].set_ylabel('Total valid cancer types')
+axarr[1].set_title('Valid cancer type count vs. PC count')
+
+sns.pointplot(data=all_data_count_df, x='n_dim', y='ratio',
+              hue='training_data', hue_order=training_data_map.values(), ax=axarr[2])
+axarr[2].set_xlabel('Number of PCs')
+axarr[2].set_ylabel('Ratio of well-predicted to total cancer type')
+axarr[2].set_title('Well-predicted/total ratio vs. PC count')
```
ben-heil, line 334: Disregard previous comment, proteome = best 'ome

Is there a reason you split out expression + methylation in their own figures, then do all data types? Are the other data types run on a subset of the samples or something?

06_predict_survival/nbconverted/plot_survival_results.py
```diff
@@ -115,75 +126,77 @@
...
 # In[9]:
 
 
 sns.set({'figure.figsize': (18, 8)})
     
-compare_df.sort_values(by='cancer_type', inplace=True)
+me_compare_df.sort_values(by='cancer_type', inplace=True)
 
-sns.boxplot(data=compare_df[~compare_df.cancer_type.isin(drop_cancer_types)],
+sns.boxplot(data=me_compare_df[~me_compare_df.cancer_type.isin(drop_cancer_types)],
```
ben-heil, line 155: Are there enough samples for a violin plot? Probably not with this many cancer types

06_predict_survival/nbconverted/plot_survival_results.py
```diff
@@ -196,12 +209,335 @@
...
-ax = plu.plot_heatmap(heatmap_df,
-                      all_results_df.reset_index(drop=True),
+ax = plu.plot_heatmap(me_heatmap_df,
+                      me_all_results_df.reset_index(drop=True),
+                      different_from_best=True,
+                      raw_results_df=raw_results_df,
+                      metric='cindex',
+                      id_name='cancer_type',
+                      scale=(-0.1, 0.4))
+
```
ben-heil, line 219: Probably not a big deal, but if, for example, the me_27k data for BLCA wasn't supposed to be in a box, it would still look like it is supposed to be since its four neighbors are 

image

ben-heil (approved):

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
```diff
@@ -0,0 +1,459 @@
...
+    ax.set_title('{} survival performance'.format(cancer_type))
+    ax.set_ylim(-0.05, 0.4)
+    if ix != 0:
+        ax.get_legend().remove()
+        
+if SAVE_FIGS:
+    plt.savefig(images_dir / 'all_top_cancers_survival.svg', bbox_inches='tight')
+    plt.savefig(images_dir / 'all_top_cancers_survival.png',
+                dpi=300, bbox_inches='tight')
+
```
nrosed, line 459: Similar to my previous comment, I don't have an intuition for why it would be decreasing here. Especially since in this plot you are looking at the difference with the shuffled labels.

06_predict_survival/nbconverted/plot_survival_results.py
```diff
@@ -196,12 +209,335 @@
...
+plt.title('Performance by cancer type for survival prediction, {}'.format(all_results_desc), pad=15)
+if SAVE_FIGS:
+    images_dir.mkdir(exist_ok=True)
+    plt.savefig(images_dir / 'survival_all_heatmap.svg', bbox_inches='tight')
+    plt.savefig(images_dir / 'survival_all_heatmap.png',
+                dpi=300, bbox_inches='tight')
+
```
nrosed, line 468: I probably missed it somewhere, but what is the difference between blue and red boxes?

nrosed:
Looks good! It's probably not necessary to fix since it's only sanity checking, but I find the QC-survival plots a bit difficult to interpret with one line per sample. It may be helpful to plot the survival curves of the learned models with the confidence intervals instead of each sample. Then for your survival plots, you will have the same number of lines as the number of conditions (age quantile, stage, etc.). It is also helpful to include the calculated hazards ratio and a p-value.  It's also nice to have tables like the one shown below to help the reader make sure the numbers per subgroup are reasonable and the p-value is reflective of a true effect.

Screen Shot 2021-09-07 at 11 10 01


Also, is the reason why LGG survival is well predicted because the subtypes are so easily captured in each of the -omic data types?

nrosed (approved):

06_predict_survival/nbconverted/plot_survival_curves.py
```diff
@@ -0,0 +1,209 @@
...
+        plt.step(fn.x, fn.y, label=sample_age_quantile,
+                 color='C{:d}'.format(sample_age_quantile))
+    except KeyError:
+        continue
+    
+handles, labels = plt.gca().get_legend_handles_labels()
+by_label = dict(zip(labels, handles))
+plt.legend(by_label.values(), by_label.keys(), title='Above/below 80th age percentile')
+plt.xlabel('Time')
+plt.ylabel('Survival probability')
```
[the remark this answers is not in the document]
jjc2718: Each line is one patient. With survival prediction, rather than predicting a single value for each sample (like you would in classification or regression) you predict a function which gives the survival probability at each time point.

So for these plots, I'm doing 4-fold cross-validation on the dataset, and for each held-out patient I'm plotting their predicted survival function and coloring the predicted functions by covariates that we think might correlate with survival.

[More detail here](https://scikit-survival.readthedocs.io/en/stable/user_guide/00-introduction.html#The-Survival-Function) if you're interested.

06_predict_survival/nbconverted/plot_survival_curves.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+        plt.step(fn.x, fn.y, label=subtype,
+                 color=sns.color_palette()[subtype_to_ix[subtype]])
+    except KeyError:
+        continue
+    
+handles, labels = plt.gca().get_legend_handles_labels()
+by_label = dict(zip(labels, handles))
+plt.legend(by_label.values(), by_label.keys(), title='{} subtype'.format(cancer_type))
+plt.xlabel('Time')
```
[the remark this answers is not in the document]
jjc2718: It is! I'll add that to the plots.

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
jjc2718: Yep, exactly right

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
jjc2718: The two separate experiments (for expression/methylation, and all data types) is the same thing we're doing for mutation prediction as well. The main reason is that in each comparison, we're using only the TCGA samples that have data for _all_ of the included data types.

That means we have about 3,000 fewer samples when we combine all data types as opposed to just looking at expression and methylation alone. So we did two separate experiments, under the assumption that we might be better powered to see some effects in the expression/methylation comparison that don't come through in the comparison of all data types.

The overall trends are pretty much the same whether we include all data types or just expression/methylation, though, so it doesn't matter too much for the main messages of the paper.

06_predict_survival/nbconverted/plot_survival_results.py
jjc2718: Yeah, you're totally right - this is something I noticed a while ago and haven't gotten around to fixing. I agree it would be better if I could move the boxes in a bit (so they would be visually more distinct for each heatmap square) but it's a bit finicky to pull it off in matplotlib.

Created issue #63 for this, hopefully I can get to it in the next week or two.

06_predict_survival/nbconverted/plot_survival_results.py
jjc2718: Yeah, there's only 8 samples per box. I looked at a violin plot here and it doesn't seem to give too much additional information.

06_predict_survival/scripts/run_survival.sh
jjc2718: Yeah! That's one of my favorite things about Bash scripting, how there's usually 10 different ways to do the same thing 😂

06_predict_survival/nbconverted/plot_survival_results.py
jjc2718: Blue = significantly better than label-permuted baseline, but significantly worse than best-performing data type
Red = significantly better than label-permuted baseline, and not significantly different from best-performing data type (i.e. "statistically equivalent to best")
No box = not significantly better than label-permuted baseline

This is in the paper already, but I'll add it to the notebook too - I don't think it's described anywhere in the code.

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
```diff
@@ -0,0 +1,459 @@
...
+axarr[1].set_xlabel('Number of PCs')
+axarr[1].set_ylabel('Total valid cancer types')
+axarr[1].set_title('Valid cancer type count vs. PC count')
+
+sns.pointplot(data=me_data_count_df, x='n_dim', y='ratio',
+              hue='training_data', hue_order=training_data_map.values(), ax=axarr[2])
+axarr[2].set_xlabel('Number of PCs')
+axarr[2].set_ylabel('Ratio of well-predicted to total cancer type')
+axarr[2].set_title('Well-predicted/total ratio vs. PC count')
+
```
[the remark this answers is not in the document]
jjc2718: Right, Casey and I don't think this plot is a very good reflection of performance at different PC counts, which is why we decided not to put it in the paper. The main reason is that it averages over cancer types - some of the cancer types have very small sample sizes so performance might jump around a lot or saturate really quickly, and when we just look at the average this results in weird trends.

When we break things out into separate cancer types, what we see makes a little more sense (at least to me), for instance these plots:

image

image

Here, it looks like either performance tends to increase a bit with added PCs (e.g. in LGG, and a bit in the pan-cancer plot) or it tends to saturate quickly and decrease (e.g. sarcoma or melanoma, maybe because the top PCs have most of the relevant info and later ones are essentially random/result in feature selection challenges). In the paper we're only going to show the pan-cancer results, and we'll put some of the individual cancer types in the supplement.

jjc2718:
> Looks good! It's probably not necessary to fix since it's only sanity checking, but I find the QC-survival plots a bit difficult to interpret with one line per sample. It may be helpful to plot the survival curves of the learned models with the confidence intervals instead of each sample. Then for your survival plots, you will have the same number of lines as the number of conditions (age quantile, stage, etc.). It is also helpful to include the calculated hazards ratio and a p-value. It's also nice to have tables like the one shown below to help the reader make sure the numbers per subgroup are reasonable and the p-value is reflective of a true effect.

Thanks! This is super helpful - we're currently not planning to use the survival curves in the paper, so while I think a figure like this would be nice, in the interest of getting the paper out soon I don't think I want to spend too much time on it now.

It is totally possible that a reviewer will ask us to show more detailed survival curves, though, so it's good to have this example in case we need to break things down more quantitatively in an actual paper figure.

> Also, is the reason why LGG survival is well predicted because the subtypes are so easily captured in each of the -omic data types?

That would be my guess. In the mutation prediction experiments we saw that predicting IDH1 status was very easy, particularly in glioma (AUROC values close to 1), so my guess is that the subtypes are captured well in the -omics data, and like you can see in the slides I linked, the subtypes are also very closely associated with survival.

06_predict_survival/nbconverted/plot_survival_pc_comparison.py
jjc2718: Yeah, I'm not entirely sure why performance seems to decrease with added PCs for some cancer types and not for others. It may have something to do with sample size - I know sarcoma and the kidney cancers tend to have fewer samples than LGG and (obviously) fewer than the pan-cancer models, so maybe the added features are causing feature selection challenges? Especially since the features here are PCs which aren't really sparse in the same sense that (for example) raw gene expression data or methylation data might be, so trying to select informative features using elastic net from a dense high-dimensional set of PCs may not be the best way to go about this.

I'll keep this in mind; maybe it would be interesting to try something similar with increasing numbers of raw features to see if the results follow the same trend.

PR #64, opened 2021-09-15
Paper updates

### LINKED ISSUE #63: Make heatmap highlights/boxes more distinct
See: https://github.com/greenelab/mpmp/pull/62#discussion_r703696830

jjc2718:
A few small updates in preparation to post a preprint:
* Reran mutational signatures experiments with standardized features, this helped performance a bit but not enough to really change the main story in the paper (mutational signatures are still the worst-performing data type)
* Moved heatmap highlights so they don't overlap (closes #63). I don't think my solution is particularly great, but it works fine and I don't want to spend a ton more time wrestling with matplotlib

Heatmap before:

all_heatmap

And after:

all_heatmap_new


ajlee21 (approved):
Code changes all look good to me, I just have a couple of comments but feel free to ignore them if you've already discussed/decided this with Casey. I know you're very close to this preprint so I don't think this should hold it up:

1. Just to check my understanding, the darker color of the cell indicates that that gene was highly predictive using data type X compared to random. Then the red indicates if the performance is significantly better than baseline AND if data type X is the "best" one compared to the others. And blue if is the performance is significantly better than baseline but the data type is NOT the best one. 
2. After seeing Milton's heatmaps, I'm wondering if annotating cells with a dot in the center will be slightly cleaner. So a large white dot indicates that the performance is significant. and then a black dot on top of the white one will indicate that in addition to the difference between significant, the data type is the "best" compared to the other methods. So something like this: 
image

3. What is your takeaway for this figure? That depending on the gene, you'll get different data types that perform well, but that expression and methylation seem to be the best ones comparatively?

jjc2718:
Thanks for this feedback, super helpful!

> 1. Just to check my understanding, the darker color of the cell indicates that that gene was highly predictive using data type X compared to random. Then the red indicates if the performance is significantly better than baseline AND if data type X is the "best" one compared to the others. And blue if is the performance is significantly better than baseline but the data type is NOT the best one.

Yeah, this is exactly right! The color of the cell is the AUPR difference vs. the baseline with permuted labels (so darker = better) and the highlights are exactly what you said.

> 2. After seeing Milton's heatmaps, I'm wondering if annotating cells with a dot in the center will be slightly cleaner. So a large white dot indicates that the performance is significant. and then a black dot on top of the white one will indicate that in addition to the difference between significant, the data type is the "best" compared to the other methods.

This is a good idea! I hadn't thought of using dots - I'll give this a try. I feel like white for significant may not be the easiest to see since there may be some low effect-size cells (close to white shading) that could be significant, but I'll play with some different dot styles/colors.

> 3. What is your takeaway for this figure? That depending on the gene, you'll get different data types that perform well, but that expression and methylation seem to be the best ones comparatively?

Yeah, I guess what you said is one of the takeaways - in general, looking at the figure you can see that most of the red shading is toward the top (expression and methylation) which supports the results from the rest of our paper showing that they tend to perform best.

Another reason I liked this figure and wanted to include it in our paper is that it summarizes specific cases nicely: for example, if you were a cancer biologist that studies EGFR, and you were designing a cell line study, you could look at this heatmap and see that gene expression is the only red cell for EGFR. That would suggest that maybe it's most important to measure gene expression in your study if you're looking for changes induced by EGFR. The rest of our figures mostly focus on overall trends (except the volcano plots, I guess, but it's pretty hard to find individual genes in those) so we thought this was a straightforward way for readers to look up whatever gene/data type they may be interested in.

jjc2718:
I talked with Casey and he finds the dots more intuitive, so we're going to stick with them. Thanks @ajlee21 for the suggestion!

PR #68, opened 2021-11-10
Fix logging file size issue

jjc2718:
I've been having an issue where for some of my long-running scripts, I end up with a log file that's 50-100 GB in size. This causes some storage issues, as you may imagine.

These changes should fix that. In short, the problem was that every time I run my scripts, the existing log file was being duplicated - with the fix I'm just appending to the existing file, resulting in much more reasonably sized (and useful) logging output.

### DIFF
--- a/02_classify_mutations/run_mutation_classification.py
+++ b/02_classify_mutations/run_mutation_classification.py
@@ -130,11 +130,6 @@ if __name__ == '__main__':
         'shuffle_labels',
         'skip_reason'
     ]
-    if io_args.log_file.exists() and io_args.log_file.is_file():
-        log_df = pd.read_csv(io_args.log_file, sep='\t')
-    else:
-        log_df = pd.DataFrame(columns=log_columns)
-        log_df.to_csv(io_args.log_file, sep='\t')
 
     tcga_data = TCGADataModel(seed=model_options.seed,
                               subset_mad_genes=model_options.subset_mad_genes,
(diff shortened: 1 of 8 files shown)


02_classify_mutations/run_mutation_classification.py
```diff
@@ -130,11 +130,6 @@ def process_args():
         'shuffle_labels',
         'skip_reason'
     ]
-    if io_args.log_file.exists() and io_args.log_file.is_file():
-        log_df = pd.read_csv(io_args.log_file, sep='\t')
-    else:
-        log_df = pd.DataFrame(columns=log_columns)
-        log_df.to_csv(io_args.log_file, sep='\t')
```
ben-heil, line 137: Just double checking to make sure I understand this change, the code used to unnecessarily read in the logs here, then overwrite them with `log_df = None` on line 157?

ben-heil (approved):
LGTM! Just making sure, the `mutation_log_df` -> `log_df` and similar changes were because you wanted to standardize names of the same object across scripts?

02_classify_mutations/run_mutation_classification.py
jjc2718: Not quite - the code here used to read in the whole existing log file, then [the code here](https://github.com/greenelab/mpmp/blob/4d77451e72339116e8acf75f23fa078ad88dbeab/mpmp/utilities/file_utilities.py#L230) would append the existing log file to the existing log file (or in other words it would duplicate it with some additional rows). Now I create a new `log_df` for each time the script is run, then I append it to the existing log file if it exists, which was my intention the first time around.

Not sure why past me thought this code was a good idea, but getting rid of it seems to achieve the original goal.

jjc2718:
> LGTM! Just making sure, the `mutation_log_df` -> `log_df` and similar changes were because you wanted to standardize names of the same object across scripts?

Yep, this was my intention!

PR #69, opened 2021-11-15
Add mutation data as features/predictors for survival

jjc2718:
We wanted to try using the true somatic mutation data for each sample as predictive features for survival, in order to compare with the predictive ability of the other -omics types (just expression and methylation for now).

For the pan-cancer dataset, this seems to perform pretty comparably to the baseline (cancer type, age, mutation burden), and worse than expression and methylation:

image

For some individual cancer types, though, the mutation data sems to work a bit better (e.g. LGG, LAML, ACC):

image


mpmp/prediction/survival.py
```diff
@@ -56,27 +65,44 @@ def train_survival(X_train,
...
+            steps=[
+                (
+                    "survival",
+                    CoxnetSurvivalAnalysis(
+                        max_iter=max_iter,
+                        tol=1e-5,
+                        # normalize input features
+                        # this seems to help with model convergence
+                        # TODO could try doing this using StandardScaler pipeline
+                        normalize=True,
```
ben-heil, line 100: I'm not familiar with Cox survival analysis. Are the results between your two versions here comparable since you normalize one but not the other?

ben-heil (approved):
I'm not really sure of what the mutation data looks like, so I don't know how to interpret the results. If it an array of "is this gene mutated in this cancer type", or a VCF, or a list of common mutations and their rates, or something else?

Like I'd expect knowing whether e.g. TP53 is mutated to be a very strong signal for survival but maybe gene expression/methylation has all that information?

jjc2718:
> I'm not really sure of what the mutation data looks like, so I don't know how to interpret the results. If it an array of "is this gene mutated in this cancer type", or a VCF, or a list of common mutations and their rates, or something else?

The features are just a binary matrix of "is this gene mutated in this sample". These come from a MAF file generated by the TCGA project. [This script here](https://github.com/greenelab/RNAseq_titration_results/blob/master/retrieve_MC3_mutations.py) is pretty close to what I've been using to generate the labels.

> Like I'd expect knowing whether e.g. TP53 is mutated to be a very strong signal for survival but maybe gene expression/methylation has all that information?

I'm not actually sure about TP53 - its mutation status is definitely a very strong signal for whether a patient has cancer or not, but conditioned on having cancer (which all of these samples do) I'm not actually sure how it affects survival.

There are some genes where the relationship is clearer, e.g. IDH mutation in gliomas which is generally [a pretty strong positive prognostic predictor](https://www.fredhutch.org/en/news/center-news/2017/05/glioma-study-links-immune-response-prognosis.html), or ERBB2 mutation in breast cancer which is a strong positive prognostic predictor because there are [effective targeted therapies](https://en.wikipedia.org/wiki/Trastuzumab) for it.

I do think a lot of the information is redundant, though, and that seems to be supported by these results (that the -omics types are generally as good or better than knowing the actual mutation status, at least for pan-cancer survival prediction).

mpmp/prediction/survival.py
jjc2718: Yeah, this is a good point - I should clean this up. Really we should be normalizing all the input features using something like `StandardScaler` (other than binary features which are probably fine not to standardize) and not using the internal `scikit-survival` normalization at all.

I don't expect this to affect the results much, if at all, though.

jjc2718: I reran these experiments using `StandardScaler` upstream in my pipeline for all non-binary covariates, and the results were essentially identical. It made the code much cleaner, though!

PR #70, opened 2021-11-18
Survival prediction using mutation predictions

jjc2718:
Following #69, we wanted to compare true mutation status as a survival predictor with _predicted_ mutation status as a survival predictor. We generated matrices of logistic regression scores for all genes in the Vogelstein dataset and all samples, and used these as real-valued features in our survival models.

For pan-cancer survival prediction, the predicted mutation status features outperform the true mutation status features pretty clearly:

image

However, for survival prediction in individual cancer types, the story isn't as clear. There are a few cancer types where the predicted mutation features help, but by and large they don't seem to provide a huge improvement.

image


02_classify_mutations/predict_all_genes.py
```diff
@@ -0,0 +1,218 @@
...
+        progress.set_description('gene: {}'.format(gene))
+
+        try:
+            tcga_data.process_data_for_gene(gene,
+                                            classification,
+                                            experiment_dir,
+                                            filter_cancer_types=False)
+        except KeyError:
+            # this might happen if the given gene isn't in the mutation data
+            # (or has a different alias)
```
ben-heil, line 155: In what cases do the genes have different aliases? Should there be a step for gene id conversion?

02_classify_mutations/predict_all_genes.py
```diff
@@ -0,0 +1,218 @@
...
+                                        model_options.num_folds,
+                                        predictor='classify',
+                                        shuffle_labels=False,
+                                        standardize_columns=False,
+                                        output_preds=True)
+            gene_preds = pd.concat(results['gene_preds'])
+            if sample_list is not None:
+                assert sample_list.equals(gene_preds.index)
+            else:
+                sample_list = gene_preds.index
```
ben-heil, line 182:
```suggestion
            # Ensure sample ids are the same across all iterations
            if sample_list is not None:
                assert sample_list.equals(gene_preds.index)
            else:
                sample_list = gene_preds.index
```

06_predict_survival/nbconverted/survival_mutations_coefs.py
```diff
@@ -0,0 +1,146 @@
...
+    'mutation_preds_me_450k': 'mutation scores, 450k',
+}
+
+
+# ### Pan-cancer models
+
+# In[4]:
+
+
+def is_feature_covariate(features):
```
ben-heil, line 56: Can you add a one line function comment to this or some markdown above the table to explain what the is_covariate column means?

06_predict_survival/nbconverted/survival_mutations_coefs.py
```diff
@@ -0,0 +1,146 @@
...
+mutation_df = pancan_data[1]
+print(mutation_df.shape)
+mutation_df.iloc[:5, :5]
+
+
+# In[11]:
+
+
+# gene = 'IDH1'
+gene = 'CARD11'
```
ben-heil, line 125: Why CARD11?

ben-heil (approved):
Do you have a hypothesis for the difference between pan-cancer and individual cancer performance? Is it that the model is better able to predict mutations given more data, so the better mutation predictions are more useful across the board? Or something else?

06_predict_survival/nbconverted/survival_mutations_coefs.py
jjc2718: Added some Markdown text about this!

jjc2718:
> Do you have a hypothesis for the difference between pan-cancer and individual cancer performance? Is it that the model is better able to predict mutations given more data, so the better mutation predictions are more useful across the board? Or something else?

Good question. I'm not completely sure, but it does seem like there's not that much difference between models in general in the pan-cancer case (x-axis above is pretty compressed). So maybe the mutation predictions are picking up on something small but predictive (like tumor stage/grade, or tumor type, etc) that doesn't help as much for the individual cancer types.

What we're hoping the mutation predictions will add is the ability identify tumors that don't have a mutation, but "act like" they have it. There are some well-known examples of this in cancer biology, including ["BRCAness" for tumors that have other mutations that emulate BRCA mutations](https://www.nature.com/articles/nrc.2015.21) and ["Ph-like" leukemias that don't have the Philadelphia chromosome translocation but share functional features and drug susceptibilities](https://ashpublications.org/blood/article/129/5/572/36151/Ph-like-acute-lymphoblastic-leukemia-a-high-risk). Unfortunately, if we were really capturing this we probably would have seen larger improvements in performance in the individual cancer types too, so it's not what we were hoping for.

PR #72, opened 2022-01-03
Batch effect correction "titration" experiments

jjc2718:
**PR description:**

As a follow-up to #71, we wanted to see what happens if we only correct a few features at a time. We did this for an increasing number of features (the "titration" part) in order to see how quickly the performance of the nonlinear models increases. The answer is "very quickly":

image

So even if only 2 or 3 features are corrected, the nonlinear models perform perfectly. This is even true for "negative control" genes where the models on the un-corrected data don't perform very well (XIST and HERC2 in the above plot). To me, all of this seems to point toward leakage of the label information into the dataset, rather than the data/problem just inherently having a lot of nonlinear signal. Pretty strange that the linear model can't pick it up, though, if it is a data leakage issue.

Interested to hear what you think!

**Summary of code changes:**

* As part of this PR, I re-implemented the `limma::removeBatchEffect` method in Python, mostly to improve my own understanding of how it works but also to speed it up a bit. I also added code to allow us to fit a limma model on the training set and apply it to the test set (that was one of our hypotheses for how the data leakage is happening -- it doesn't seem to make any difference, unfortunately). This code is in `mpmp/utilities/batch_utilities.py`.
* Analysis code is mainly in `02_classify_mutations/batch_effects/bc_titration.ipynb`, with some small changes to `02_classify_mutations/batch_effects/batch_correction.ipynb` (these don't need a close review).
* `02_classify_mutations/batch_effects/batch_correction_titration.py` is mostly a copy of `02_classify_mutations/run_mutation_classification.py`, with some small changes. This probably doesn't need a super close review.
* Scripts in `02_classify_mutations/batch_effects/scripts` are all pretty similar to existing code and don't need close review.


mpmp/utilities/batch_utilities.py
```diff
@@ -5,52 +5,129 @@
...
+    design = contrast.matrix[batches.astype(int), :]
+
+    # if coefficients are provided, just use them to correct the provided data
+    # otherwise fit the model and correct the provided data
+    if coefs is None:
+        from sklearn.linear_model import LinearRegression
+        # X is an n x p matrix
+        # batches is a n x m vector of batch indicators
+        # we want to find a m x p vector of coefficients
+        reg = LinearRegression().fit(design, X)
```
ben-heil, line 85: Double checking: limma uses the design matrix to predict the data rather than just predicting the class from the labels and subtracting the residuals?

ben-heil: That is to say that X is the dependent variable and design is the independent?

02_classify_mutations/batch_effects/bc_titration.ipynb
```diff
@@ -0,0 +1,528 @@
...
+    "                ]\n",
+    "                handles, labels = ax.get_legend_handles_labels()\n",
+    "                l = ax.legend(\n",
+    "                    [Line2D([0], [0], color='black', linestyle='--')] + handles +[Line2D([0], [0], color='black', linestyle='--')] + custom_legend_lines,\n",
+    "                    ['model'] + ['linear', 'nonlinear'] + ['signal'] + ['signal', 'shuffled']\n",
+    "                , ncol=2)\n",
+    "                l.legendHandles[0].set_visible(False)\n",
+    "                l.legendHandles[3].set_visible(False)\n",
+    "            else:\n",
+    "                ax.legend().set_visible(False)\n",
```
ben-heil: It's interesting to me that different problems take different numbers of features to reach 100 percent accuracy. I wonder whether that's a function of the process (some features are more corrupted by be correction) or the data (some problems have something that makes it hard to learn useful signal). If you're looking for a positive control to run, you could titrate in setting X number of features to the true labels and see how long it takes the model to hit 100% accuracy

ben-heil (approved):
> To me, all of this seems to point toward leakage of the label information into the dataset, rather than the data/problem just inherently having a lot of nonlinear signal. Pretty strange that the linear model can't pick it up, though, if it is a data leakage issue.

I agree, it definitely looks like data leakage. I'm not surprised that linear models aren't able to pick up the results though. Imagine you're trying to tell the difference between two distributions with different amounts of skew. If you were to set both their means to zero, a model that only looks at means wouldn't be able to tell them apart. However, a nonlinear model that isn't solely dependent on the mean could tell them apart pretty easily. 
image

I think the metaphor is a little bit strained though because I believe the correction is causing nonlinear signals the data. A person looking to do differential expression wouldn't care (I think) because they're trying to estimate a linear quantity anyway. I'm currently playing around with some toy examples to get an intuition for how sum coding works, so will have more insight SOON (soon or otherwise never)

mpmp/utilities/batch_utilities.py
jjc2718: > That is to say that X is the dependent variable and design is the independent?

Yes, that's right. The idea is to predict the original data matrix as a linear function of the batch information, then subtract away that approximation to get the corrected/adjusted data matrix. This idea is described [in this paper](https://www.biorxiv.org/content/10.1101/2020.08.17.255034v1.full.pdf) (see the "Problems of input adjustment" section), and the R code from limma is [here](https://www.biostars.org/p/434736/) which should do the same thing (although they don't explicitly describe it anywhere as far as I know).

> just predicting the class from the labels and subtracting the residuals

How would this work? It sounds kind of similar to what I'm doing but I'm not sure I totally follow.

02_classify_mutations/batch_effects/bc_titration.ipynb
jjc2718: This is an interesting idea! My guess is that the positive control would hit 100% accuracy pretty quickly but it would be good to try.

jjc2718:
> I agree, it definitely looks like data leakage. I'm not surprised that linear models aren't able to pick up the results though. Imagine you're trying to tell the difference between two distributions with different amounts of skew. If you were to set both their means to zero, a model that only looks at means wouldn't be able to tell them apart. However, a nonlinear model that isn't solely dependent on the mean could tell them apart pretty easily.
> 
> I think the metaphor is a little bit strained though because I believe the correction is causing nonlinear signals the data. A person looking to do differential expression wouldn't care (I think) because they're trying to estimate a linear quantity anyway. 

Makes sense to me! I wonder why your results weren't as good/near perfect after you batch correct out your tissue signal, though (I think I remember you tried the same thing). If I remember correctly your nonlinear models did better than the linear ones, but the performance wasn't suspiciously good.

> I'm currently playing around with some toy examples to get an intuition for how sum coding works, so will have more insight SOON (soon or otherwise never)

Interested to see what you find (if you get to it)! This is quite the mystery, my intuition has been totally wrong almost every step of the way 😂

mpmp/utilities/batch_utilities.py
jjc2718: But then the residuals would be in Y-space, right? Or in other words, there would just be a single residual for the prediction of Y, rather than one for each column in X - would you just subtract the same residual from each column in X?

I guess these are similar ideas, it just makes more intuitive sense to me to have a different prediction/residual/correction for each column of X. I'm not super convinced that it's "right", though, I think there's probably lots of ways to do this.

ben-heil: > But then the residuals would be in Y-space, right? Or in other words, there would just be a single residual for the prediction of Y, rather than one for each column in X - would you just subtract the same residual from each column in X?
> 
> I guess these are similar ideas, it just makes more intuitive sense to me to have a different prediction/residual/correction for each column of X. I'm not super convinced that it's "right", though, I think there's probably lots of ways to do this.

After looking up how statistics works, I think it would be roughly the same idea in the other direction. Instead of subtracting out the signal from the data you'd subtract out the linear prediction from the labels. What remains would be the parts of the label unpredictable via linear model.

It's less intuitive, but it works for resnets so should be at least possible (though not necessarily helpful)

jjc2718: > you'd subtract out the linear prediction from the labels

Ahh, got it -- so you'd be modifying the labels rather than the data. That does seem like it would have a similar effect (as far as removing linear signal) - maybe worth trying in the future.

PR #73, opened 2022-02-09
Feature selection using f-test

jjc2718:
We wanted to try feature selection for our models using univariate correlation with the labels (on the training set), which we implemented using `SelectKBest` and `f_classif` from scikit-learn.

For a small subset of genes we tried, it doesn't seem to affect performance much as compared to selecting features by top mean absolute deviation:

image


ben-heil (approved):
Looks good! Have you compared MAD/f-test feature selection to using random features? I'd expect something smart to do better than random but it may be worth checking

jjc2718:
> Looks good! Have you compared MAD/f-test feature selection to using random features? I'd expect something smart to do better than random but it may be worth checking

Good idea! I'll try that. I would expect it to perform worse, but it would be a good baseline.

> Also, does your conda install usually take more than six hours or is there a bug?

I think there's a bug - I've installed this conda env on a few new systems lately and it usually only takes a few mins at most. I ran the tests locally and they all passed.

jjc2718:
Results with random feature selection:

image

Interesting that it's not too bad for some genes! My guess is that those genes are ones that rely really heavily on the non-gene expression covariates (i.e. cancer type and mutation burden). The random features seem to be uniformly worse than the other methods, though, so that makes sense.

PR #74, opened 2022-02-13
Explore gene dosage effects

jjc2718:
For our submitted paper, one of the reviewers asked how important the expression levels of the mutated gene (i.e. "gene dosage") are in our predictions: see greenelab/mpmp-manuscript#45.

The answer, for a limited set of genes, seems to be "generally not that important" (green and orange boxes similar and green and purple boxes different):

image

See the notebook `02_classify_mutations/dosage_effects/plot_dosage_effects.ipynb` for additional details.


mpmp/data_models/tcga_data_model.py
```diff
@@ -227,6 +231,17 @@ def process_data_for_gene(self,
                 np.array([cancer_type_to_index[ct] for ct in y_filtered_df.DISEASE]),
                 verbose=self.verbose)
 
+        print(train_filtered_df.shape)
+        print(gene_features.shape)
+
```
ben-heil, line 236: These are probably no longer necessary (though I may be wrong)
```suggestion

```

mpmp/utilities/tcga_utilities.py
```diff
@@ -829,4 +829,74 @@ def compress_and_save_data(data_type,
...
+    gene_features_updated = np.array(
+        [True] + ([False] * non_gene_df.shape[1])
+    )
+
+    return data_df_updated, gene_features_updated
+
+
+def get_symbol_map():
+
+    # TODO move these urls to config
```
ben-heil, line 883: Is this still in the cards, or is it alright for the URL to live here?

ben-heil (approved):
Looks good! I wonder why gene dosage is relevant in some genes and not in others. Maybe they're more isolated in the coexpression network, so other genes' values aren't able to compensate for them as easily?

mpmp/utilities/tcga_utilities.py
jjc2718: Still in the cards! Good catch, will do this now.

jjc2718:
> Looks good! I wonder why gene dosage is relevant in some genes and not in others. Maybe they're more isolated in the coexpression network, so other genes' values aren't able to compensate for them as easily?

Yeah, it could be something like that. EGFR and ERBB2 are both cell surface receptors that are relatively "upstream" in their cancer-associated pathways, so maybe they have more unique information about cell signaling/tumor state etc. while more "downstream" genes (the rest of them AFAIK) are more redundant with the rest of their cancer pathway. I dunno, I'll have to think through it more.

PR #75, opened 2022-02-19
COSMIC CGC gene set analysis, part 1

jjc2718:
**PR description:**

For our paper in review, both reviewers were interested in seeing results for a larger set of cancer genes (greenelab/mpmp-manuscript#43),  with one of them pointing out the [COSMIC Cancer Gene Census](https://cancer.sanger.ac.uk/cosmic/census?tier=1) and its larger set of DNA damage repair genes compared to the Vogelstein et al. gene set we've been using.

This PR lays the foundation for a more in-depth analysis of this gene set. So far I've only run the "all data types" comparison for these genes, and I'll likely rerun all of our analyses in a future PR and combine this gene set with the Vogelstein genes based on the results here.

With the Vogelstein gene set, which is considerably smaller, we saw generally similar performance for expression and methylation. When we expand to the COSMIC gene set, this seems to favor gene expression:

image

And counting the number of "well-predicted" genes at a p-value cutoff of 0.001 (these numbers were pretty similar between data types for the Vogelstein genes):

Screen Shot 2022-02-19 at 4 16 53 PM

**Code changes:**

* Added `01_explore_data/explore_cosmic_gene_set.ipynb` to download and explore the COSMIC CGC genes
* Added code to load oncogene/TSG information for COSMIC genes and run our scripts on them
* Visualized results in `02_classify_mutations/plot_all_results.ipynb` (worked with minimal changes)
* Added some things that I've been using on the cluster that aren't tracked in git (`02_classify_mutations/scripts/slurm_scripts/run_drop_target_vogelstein.sbatch`, `mpmp/scripts/check_complete.py`) - these don't really need to be reviewed closely


01_explore_data/nbconverted/explore_cosmic_gene_set.py
```diff
@@ -0,0 +1,235 @@
...
+class_df = pd.read_excel(
+    cfg.data_dir / '1-s2.0-S009286741830237X-mmc1.xlsx', 
+    engine='openpyxl', sheet_name='Table S1', index_col='KEY', header=3
+)
+class_df.rename(columns={'Tumor suppressor or oncogene prediction (by 20/20+)':
+                         'classification'},
+                inplace=True)
+
+print(class_df.shape)
+class_df.head()
```
ben-heil, line 83: Looking at the resulting dataframe here, there are a lot of unnamed columns with largely NaN values. Do you know why that is? Just making sure there wasn't a weird delimeter/string escaping issue that might corrupt some rows

01_explore_data/nbconverted/explore_cosmic_gene_set.py
```diff
@@ -0,0 +1,235 @@
...
+print(cosmic_df['Role in Cancer'].unique())
+
+
+# In[5]:
+
+
+# load Bailey et al. data
+# supplementary table from https://www.sciencedirect.com/science/article/pii/S009286741830237X
+# this contains oncogene/TSG predictions for genes/cancer types using 20/20+ classifier
+class_df = pd.read_excel(
```
ben-heil, line 74: Is there an efficient way to make sure that none of the genes got coerced into dates due to exposure to Excel?

ben-heil (approved):
Not saying you should do this, especially not just to address reviewers' comments, but it would be interesting to see the extent to which COSMIC genes do better due to more cancer information vs due to just having more genes

01_explore_data/nbconverted/explore_cosmic_gene_set.py
jjc2718: Good catch - these are all empty columns in the original Excel file. I think I can just remove them.

PR #76, opened 2022-02-23
Add script to count valid samples/cancer types

jjc2718:
In the preprocessing code for our classifiers, we filter out cancer types that don't contain at least 5% of samples mutated and at least 10 total samples mutated, for a given target gene.

We were curious how many total genes these filters would give us, if we look at all ~20,000 genes we have mutation data for. This script filters samples for each gene and counts the number of samples/cancer types that would be included in our classifiers.

There are quite a few valid genes:

Screen Shot 2022-02-23 at 12 55 59 PM

(when CNV data is included)

Screen Shot 2022-02-23 at 12 58 10 PM

(when we just look at point mutations)


01_explore_data/nbconverted/count_all_genes_filters.py
```diff
@@ -0,0 +1,193 @@
...
+# load sample info and mutation data, this takes some time
+tcga_data = TCGADataModel(seed=cfg.default_seed, verbose=False)
+pancancer_data = du.load_pancancer_data()
+
+
+# In[4]:
+
+
+def gene_sample_count(gene, data_model, classification='neither'):
+    """Count valid samples/cancer types for a given gene."""
```
arielah, line 56: What do you mean, count cancer types for a given gene? Number of cancer types where that gene is implicated?

01_explore_data/nbconverted/count_all_genes_filters.py
```diff
@@ -0,0 +1,193 @@
...
+        
+    return (gene, sample_count, cancer_types)
+
+
+# In[5]:
+
+
+# cache partial results and load them
+# we're running this script for 20,000 genes, so it's nice to save progress in case
+# execution is interrupted
```
arielah, line 78: Is there a scenario where execution fails partway through a gene and so some information for that single gene gets written but not all?

arielah (approved):
Looks good! Just a couple of clarification questions.

01_explore_data/nbconverted/count_all_genes_filters.py
jjc2718: Number of cancer types where we have enough mutations in that gene (5% and at least 10 total mutated samples) to train a classifier.

I'll clarify this in the comment!

01_explore_data/nbconverted/count_all_genes_filters.py
jjc2718: I don't think that should happen - in [the code here](https://github.com/greenelab/mpmp/blob/46b5549edf8a1a6e5499d86050fff5dc70f2756d/01_explore_data/nbconverted/count_all_genes_filters.py#L110), I'm only saving the information every `save_every` iterations, so if execution fails partway through a gene we'd just go back to the previous iteration. We might have to rerun a few genes that way, but that's fine assuming `save_every` is fairly small (on the order of tens or hundreds).

PR #79, opened 2022-02-24
COSMIC CGC gene set analysis, part 2

jjc2718:
Based on the results from #75 , we decided to generate a new, larger gene set that includes the genes from both COSMIC and Vogelstein et al. We also included genes from [Bailey et al. 2018](https://www.sciencedirect.com/science/article/pii/S009286741830237X), in an effort to make this a fairly comprehensive set of cancer genes.

Results of rerunning the scripts for this larger gene set are fairly similar to what we saw for the COSMIC genes in #75, with expression again coming out the clear winner:

image

Screen Shot 2022-02-22 at 3 17 42 PM

Screen Shot 2022-02-22 at 3 18 08 PM

**Code changes:**

* Created `00_download_data/2_download_cancer_gene_data.ipynb` to download and merge cancer gene information
* Added code to `mpmp/utilities/data_utilities.py` to load merged gene set
* Reran scripts in `02_classify_mutations` to plot results on merged gene set


00_download_data/nbconverted/2_download_cancer_gene_data.py
```diff
@@ -0,0 +1,290 @@
...
+
+print(vogelstein_bailey_df.classification.unique())
+vogelstein_bailey_df.head()
+
+
+# In[11]:
+
+
+# examples where the datasets disagree
+vogelstein_bailey_df[vogelstein_bailey_df.classification == 'check'].head()
```
ben-heil, line 195: Very minor point, but I thought you were fixing only the first four discordant genes manually rather than all the discordant genes due to the `.head()` call here

mpmp/config.py
```diff
@@ -182,8 +182,20 @@
...
+bailey_raw_file = cancer_genes_raw_dir / '1-s2.0-S009286741830237X-mmc1.xlsx'
+
+# location of annotated cancer gene info
+# these annotations are collated/cleaned up in 2_download_cancer_genes.ipynb
+cosmic_with_annotations = cancer_genes_dir / 'cosmic_tier1_annotated.tsv'
+bailey_with_annotations = cancer_genes_dir / 'bailey_annotated.tsv'
+vogelstein_with_annotations = cancer_genes_dir / 'vogelstein_annotated.tsv'
+merged_cancer_genes = cancer_genes_dir / 'merged_with_annotations.tsv'
+
+# TODO: move?
```
ben-heil, line 198: Double checking that this TODO should still live here

ben-heil (approved):
Transcriptomics = best omics

00_download_data/nbconverted/2_download_cancer_gene_data.py
jjc2718: Ahh, got it. I guess I don't need the `head()` call since there's only four of them total.

mpmp/config.py
jjc2718: Nope, I did move it and I forgot to delete the TODO and the variable assignment on the next line. I'll fix this now.

PR #80, opened 2022-02-25
Run "drop target" experiment for all Vogelstein genes

jjc2718:
Following on #74, we decided to run the "drop target" experiments (dropping the target gene from the gene expression dataset) for all of the Vogelstein genes.

Similar to our previous results, we saw that it makes very little difference for every gene in this gene set, and none of the differences are significant in a t-test:

image

This isn't too surprising given what we know about gene expression data and the redundancy/correlation between features, but it's good to confirm that the expression of the mutated gene isn't the main driver of good performance in most cases.


ajlee21 (approved):
The code looks good to me. Just one clarifying question:

So previously if your labels were mutations in gene A, you would still use expression data from gene A in your training/input data? But here you're removing the gene A column from your expression data? And you see the predictiveness of gene A mutations are unchanged due to probably some other redundant signals in the input?

Good experiment to assess the contribution of genes. If its due to redundant information then i would assume that using PC features and then removing the one containing/gene A is high weight in might reduce your predictive signal then? Or maybe other complementary gene sets (non gene A PCs are contributing). Just a thinking out loud. But I haven't spent as much time thinking about this project as you so perhaps not a worth while experiment.

jjc2718:
> So previously if your labels were mutations in gene A, you would still use expression data from gene A in your training/input data? But here you're removing the gene A column from your expression data? And you see the predictiveness of gene A mutations are unchanged due to probably some other redundant signals in the input?

Yep, exactly!

> Good experiment to assess the contribution of genes. If its due to redundant information then i would assume that using PC features and then removing the one containing/gene A is high weight in might reduce your predictive signal then? Or maybe other complementary gene sets (non gene A PCs are contributing). Just a thinking out loud. But I haven't spent as much time thinking about this project as you so perhaps not a worth while experiment.

Yeah, I definitely think there are other things we could try, like removing the entire pathway the gene is in or removing PCs like you suggested. For now we're doing this primarily to address a reviewer's comment, and they just asked about the target gene specifically, so I'll move on for now but keep these ideas in mind.

I do remember that Greg tried something like this in [the Ras pathway paper](https://pubmed.ncbi.nlm.nih.gov/29617658/) - they tried dropping all "RASopathy" genes from the classifier (a broader set of genes involved in the Ras/MAPK pathway) and they found that it didn't affect performance. So I do think there's a lot of redundant signal in the data for sure, but it could be interesting to think about the limits of that at some point.

PR #81, opened 2022-02-28
Extend top/random gene sets to match size of merged gene set

jjc2718:
In #79, we changed most of the single-omics results from our paper to use a new, larger set of cancer genes formed by "merging" a few different gene sets (rather than just the Vogelstein et al. data we were looking at before).

This PR changes the numbers of top mutated/random genes to match the size of the new gene set. The results are still more or less the same, in the sense that choosing genes for their involvement in cancer still results in more well-performing classifiers than choosing genes by mutation frequency or randomly.

Screen Shot 2022-02-28 at 11 24 32 AM

image


arielah (approved):
Looks good!

PR #82, opened 2022-03-02
GO functional enrichment analysis of cancer gene sets

jjc2718:
In #79, we saw that when we switch from using the Vogelstein et al. cancer genes to a larger set of cancer-associated genes, relative performance between data types shifts toward favoring gene expression over other data layers.

We wanted to explore why this might be the case by looking at GO functional enrichment of the different datasets. This is implemented in `01_explore_data/explore_cancer_gene_sets.ipynb`, and the results are summarized at the end of that script.

TL;DR: nothing jumps out at me as being a super obvious difference between the datasets, but the non-Vogelstein cancer genes might be enriched for transcription factors/genes involved in transcription, which could be reflected in the usefulness of the gene expression data for the added genes. The newly added genes also seem to include slightly more cell cycle regulators and genes involved in DNA binding.


01_explore_data/nbconverted/explore_cancer_gene_sets.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## Explore cancer gene sets
+# 
+# In `00_download_data/2_download_cancer_gene_data.ipynb`, we downloaded and integrated several different sets of cancer-associated genes. In this script, we want to explore the similarities and differences between these gene sets in more detail.
+# 
+# We see that the results of our classification experiments are somewhat dependent on which gene set or sets we decide to use, so understanding the similarities and differences between the gene sets will help us to put these in context.
```
ajlee21, line 8: Maybe helpful to link to the different publication/DB sources where you're downloading the data from.

Also, I assume the motivation for looking at these different input datasets is to see if your results/prediction trends generalize. Is that correct

01_explore_data/nbconverted/explore_cancer_gene_sets.py
```diff
@@ -0,0 +1,388 @@
...
+
+vogelstein_genes = set(vogelstein_df.gene.values)
+cosmic_genes = set(cosmic_df.gene.values)
+bailey_genes = set(bailey_df.Gene.values)
+
+
+# In[6]:
+
+
+from venn import venn
```
ajlee21, line 80: Nice plot.

I assume you've already checked this, but I know that gene id mapping can always be tricky. So you made sure that the differences are not due to some mapping issue?

01_explore_data/nbconverted/explore_cancer_gene_sets.py
ajlee21: Maybe helpful to add a short description of the different gene sets - i.e. number of total genes, what type of patients and how they determined oncogene vs non-oncogene

01_explore_data/nbconverted/explore_cancer_gene_sets.py
```diff
@@ -0,0 +1,388 @@
...
+label_map = {
+    'cosmic': cosmic_genes,
+    'bailey': bailey_genes,
+    'vogelstein': vogelstein_genes
+}
+venn(label_map)
+plt.title('Overlap between cancer gene sets', size=13)
+
+
+# ### Enrichment analysis of gene sets
```
ajlee21, line 95: It seems like there is a lot of processing happening in these cells that could organizationally could be split out into a separate nb to be cleaner. But I completely understand if this is a one-off/quick analysis that you wanted to keep everything together. So completely up to you, just something that I noticed.

Also, I'm not as familiar with ontologies and the format of them, so there are many steps here that I'm not exactly sure why you're doing it. Not sure if this can easily read if you provide an example of what the ontology looks like?

01_explore_data/nbconverted/explore_cancer_gene_sets.py
```diff
@@ -0,0 +1,388 @@
...
+
+vogelstein_output_file = cfg.go_output_dir / 'vogelstein_enrichment.tsv'
+goea.wr_tsv(str(vogelstein_output_file), vogelstein_sig_results)
+
+
+# In[18]:
+
+
+# get non-Vogelstein gene set
+all_genes = set.union(*list(label_map.values()))
```
ajlee21, line 252: Just checking that non-Vogelstein = (Bailey + COSMIC) - Vogelstein?

ajlee21 (approved):
Really really interesting experiment. This was super neat to look at.

I wonder if the genes in Vogelstein are picking up more secondary expression effects vs this large non-Vogelstein set that contains core regulators that have such a strong impact on expression. 

Did you find that all non-Vogelstein genes are more predictive using expression? Or just that there are some that are compared to using Vogelstein?

Do you think it its how the oncogenes are defined in the different datasets?

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: I do have most of this information in the script at `00_download_data/2_download_cancer_gene_data.ipynb` where we actually download the data - I'll make it clear here that there's more information about the gene sets in that script.

> Also, I assume the motivation for looking at these different input datasets is to see if your results/prediction trends generalize. Is that correct

Yep! Both reviewers on our paper wanted to see if our conclusions generalized outside of the Vogelstein dataset, particularly to a dataset with more DNA damage repair genes (which COSMIC CGC has a lot of). Turns out it does change our conclusions a bit, so it was a helpful review comment!

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: Yep, exactly! It's the combination of the two non-Vogelstein datasets, minus the Vogelstein genes that overlap with those datasets.

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: By "gene id mapping", do you mean the mapping of genes to GO terms? Or the mapping of gene symbols to Entrez IDs?

We're doing both in this script - just want to check which you've found tricky (or if both are tricky). For the gene -> GO term mapping I basically just followed a tutorial, and the symbol -> Entrez mapping I've done several times before so I'm a bit more familiar.

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: Yeah, this is definitely not as clean as it could be. I think the goal is just a quick one-off analysis, so I may leave it how it is, but I'll try to document it a bit better.

I do think it's possible to plot the ontology (or part of it) using the `goatools` package, so I'll look into that.

01_explore_data/nbconverted/explore_cancer_gene_sets.py
ajlee21: Sorry I was referring to the gene symbol to entrez not always being 1-1. But I'm sure you checked I was double checking just in case

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: I added an example of plotting part of the ontology that's relevant to the non-Vogelstein genes at the end of the script. Here's what it looks like:

image

GO terms and the relations between them generally look like this (a directed acyclic graph, or DAG). Plotting the whole thing could get super gnarly, so I had to cherry-pick a few terms that we're interested in.

01_explore_data/nbconverted/explore_cancer_gene_sets.py
jjc2718: I haven't checked super closely, I'm relying here on the gene mapping that was developed for Cognoma (see, e.g. [this PR](https://github.com/cognoma/cancer-data/pull/12)) which Greg has also used in most of his classification repos. As far as I can tell they dropped ambiguous symbol -> ID mappings from their maps, and almost all of the genes here seem to map to IDs, so I think I'm okay with that since this is a quick one-off analysis.

I don't think the differences in enriched GO terms that we see here are due to gene mapping ambiguities -- I think they're almost certainly due to actual differences in the genes in each gene set -- but it's still worth thinking about!

jjc2718:
> Did you find that all non-Vogelstein genes are more predictive using expression? Or just that there are some that are compared to using Vogelstein?

We found that proportionally more genes are better predicted using expression in the larger gene set than in the Vogelstein gene set alone. It's definitely not _all_ of the new genes that are better predicted with expression, but a lot of them seem to be.

> Do you think it its how the oncogenes are defined in the different datasets?

I don't think so - we only use oncogene/TSG info to determine whether/which kind of CNV info to use to determine mutated samples. So even if we get the wrong annotation, it would just add a small amount of noise into our labels.

PR #83, opened 2022-03-14
Add Bayesian optimization for multi-omics hyperparameter selection

jjc2718:
**PR description:**

Sorry in advance, this is a bit of a large PR with a few interdependent changes. One of our reviewers asked us about our parameter ranges/model fits for the multi-omics experiments, and we wanted to do our best to rule out the parameter ranges as a potential explanation for the comparative performance of single- and multi-omics models.

This PR adds Bayesian optimization, as implemented in [`scikit-optimize`](https://scikit-optimize.github.io/stable/index.html), as an option for hyperparameter selection. We ran this for 100 iterations (instead of our previous grid search) for the single- and multi-omics models, and did the same comparison as before using the same genes.

As expected, we see that Bayesian optimization generally provides a slight increase in performance across the board, relative to grid search:

image

However, this doesn't seem to change the _relative_ performance substantially between the single- and multi-omics models (after statistical testing/multiple testing correction, we still see that none of the differences are significant, although maybe one could argue that SETD2 is "trending toward significance"):

image

**Main code changes:**

* Implementation/application of Bayesian optimization is in `mpmp/prediction/classification.py`
* Comparison notebook added in `05_classify_mutations_multimodal/compare_bayes_opt.ipynb`
* Added a function to `mpmp/prediction/cross_validation.py` to run a single fold of cross-validation (this allows us to run the hyperparameter search for each CV fold in parallel, on the cluster)
* Add the ability to write the hyperparameter search results to the results directory (this is mostly in `mpmp/prediction/cross_validation.py`)


05_classify_mutations_multimodal/nbconverted/compare_bayes_opt.py
```diff
@@ -0,0 +1,195 @@
...
+
+# In[5]:
+
+
+grid_results_df = pd.concat((
+    grid_results_df, u_results_df
+))
+group_df = (grid_results_df
+    .groupby(['identifier', 'training_data'])
+    .count()
```
ben-heil, line 91: Depending on what you're looking for, it may be more helpful to call `unique` after `count`. In theory only one row should show up, indicating that all the fields for all the data types have the same counts. If that's not the case it should show you which data types/genes are different

mpmp/prediction/classification.py
```diff
@@ -69,19 +69,110 @@ def train_classifier(X_train,
...
+                    max_iter=cl_max_iter,
+                    tol=1e-3,
+                ),
+            )
+        ]
+    )
+
+    search_spaces = {
+        'classify__alpha': (1e-6, 1e6, 'log-uniform'),
+        'classify__l1_ratio': (0.0, 1.0, 'uniform')
```
ben-heil, line 154: Are the double underscores in these variables an `skopt` thing or something you came up with? I've never seen double underscores within a variable name before

mpmp/prediction/cross_validation.py
```diff
@@ -337,6 +365,191 @@ def run_cv_stratified(data_model,
...
+        # (i.e. preserve the total number of mutated samples in each)
+        original_ones = y_train_df.groupby('DISEASE').sum()['status']
+        y_train_df.status = shuffle_by_cancer_type(y_train_df, data_model.seed)
+        y_test_df.status = shuffle_by_cancer_type(y_test_df, data_model.seed)
+        new_ones = y_train_df.groupby('DISEASE').sum()['status']
+
+        # number of mutated samples per cancer type should be the same before
+        # and after shuffling
+        assert original_ones.equals(new_ones)
+
```
ben-heil, line 432: Triple newline should be single newline for PEP reasons (sorry I can't figure out how to remove multiple newlines in a single suggestion).
```suggestion
```

mpmp/prediction/cross_validation.py
```diff
@@ -337,6 +365,191 @@ def run_cv_stratified(data_model,
...
+
```
ben-heil, line 433:
```suggestion
```

mpmp/prediction/cross_validation.py
```diff
@@ -337,6 +365,191 @@ def run_cv_stratified(data_model,
...
+                num_folds,
+                fold_no,
+                shuffle_labels=False,
+                standardize_columns=False,
+                num_features=-1,
+                feature_selection_method='mad',
+                bayes_opt=False,
+                nonlinear=False,
+                output_preds=False,
+                output_grid=False):
```
ben-heil, line 382: This function is missing a docstring

ben-heil (approved):
So what you're saying is the results are Bayesically the same?

02_classify_mutations/run_mutation_compressed.py
```diff
@@ -61,6 +63,11 @@ def process_args():
                                      'parameters for training/evaluating model, '
                                      'these will affect output and are saved as '
                                      'experiment metadata ')
+    opts.add_argument('--bayes_opt', action='store_true',
```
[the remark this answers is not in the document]
jjc2718: Yeah, I agree - having to add these to multiple different scripts is a bit of a mess. I will definitely try to make the argument parsing more modular in the future, but like you said, at this point it probably doesn't make sense for this project.

mpmp/prediction/classification.py
jjc2718: They're actually an [sklearn `Pipeline` thing](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html), which `skopt` emulates. The classifier is one step of the pipeline (in this case the only step), so its arguments are prefixed with the step name, followed with the double underscore and the parameter name.

05_classify_mutations_multimodal/nbconverted/compare_bayes_opt.py
jjc2718: Good idea! That's a much more concise way to find the rows that are different.

05_classify_mutations_multimodal/nbconverted/compare_bayes_opt.py
```diff
@@ -0,0 +1,195 @@
...
+                order=data_order, hue='param_opt', ax=ax)
+    ax.set_title('Prediction for {} mutation'.format(gene))
+    ax.set_ylabel('AUPR')
+    ax.set_ylim(0.0, 1.1)
+    for tick in ax.get_xticklabels():
+        tick.set_rotation(40)
+    if ix != 0:
+        ax.legend_.remove()
+        
+plt.tight_layout()
```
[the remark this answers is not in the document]
jjc2718: Yeah, good point. This plot is showing the variance across cross-validation folds (each box plot is a classifier trained using a single data type for a single gene, across 8 train/test splits). So maybe it's possible that a better choice of hyperparameters more accurately reflects the performance variability between folds? Dunno.

I'm not too worried about it, though - as long as the results generally tend to be better than the grid search results I'm fairly confident that it's working well.

cgreene:
> So what you're saying is the results are Bayesically the same?

😂 😂 😂 😂 😂 😂 😂 😂 😂 😂 😂 😂

PR #85, opened 2022-03-22
Gene-level filtering vs. per-cancer type filtering

jjc2718:
Currently, to build the training/testing dataset for our classifiers, we filter to cancer types that have at least 5% of samples mutated and at least 10 total samples mutated. One of our paper reviewers brought up the idea that instead of applying these filters to samples from each cancer type independently, we could instead just look at the whole dataset - or in other words, we could choose genes to train classifiers for based on their _overall_ percent/count of mutated samples.

This PR implements this alternate gene-level filtering method and compares the results with our existing classifiers. Overall, the results are fairly similar in the sense that gene expression still seems to perform considerably better than the other data types (as we saw in #75 and #79), and in general performance tends to be worse with the gene-level filtering approach.

Here is a plot showing the difference for each gene between the "old" filtering scheme (per-cancer type) and the "new" filtering scheme (gene-level across all cancer types). A positive value for a gene means that gene's classifier performed better for the "old" filtering scheme and vice-versa for the "new" filtering scheme:

image

So we can see that most values are above 0 here, indicating that filtering for each cancer type independently tends to lead to better classifier performance for most genes. This seems to hold across all the data types we looked at.

**Code changes:**

* `01_explore_data/count_dataset_filters.ipynb` tests some different mutation percentage/mutation count filters (I settled on 100 total mutations and 1% of samples mutated across all cancer types, which seems to be a decent tradeoff between sensitivity and specificity)
* `02_classify_mutations/compare_filtering.ipynb` compares the "old" and "new" filtering schemes
* `mpmp/utilities/tcga_utilities.py` has the implementation of the gene-level filtering


02_classify_mutations/nbconverted/compare_filtering.py
```diff
@@ -0,0 +1,300 @@
...
+sns.set({'figure.figsize': (22, 5)})
+sns.set_style('whitegrid')
+
+fig, axarr = plt.subplots(1, 3)
+
+# plot mutation prediction from expression, in a volcano-like plot
+datasets = ['gene expression', '27k methylation', '450k methylation']
+filtered_data_map = {k: v for k, v in training_data_map.items() if v in datasets}
+
+plu.plot_volcano_baseline(new_all_results_df,
```
arielah, line 124: Is it weird that you get fewer significant genes with 450k methylation than with 27k methylation? Based on your previous work this surprised me, so I'm wondering if I read this figure wrong.

02_classify_mutations/nbconverted/compare_filtering.py
```diff
@@ -0,0 +1,300 @@
...
+                          verbose=True)
+
+
+# ### Compare against results with old filtering scheme
+
+# In[9]:
+
+
+# load raw data
+old_results_df = au.load_stratified_prediction_results(orig_results_dir, 'gene')
```
arielah, line 158: Is it worth it to change the names from "old" and "new" to something like "total_filtering" vs. "cancerwise_filtering"? If someone in the future is reading this code doesn't know what order you did things in, this notation might be confusing. But it's not too hard to figure out, so your call.

02_classify_mutations/nbconverted/compare_filtering.py
```diff
@@ -0,0 +1,300 @@
...
+        
+show_values_on_bars(plt.gca())
+plt.gca().get_xaxis().set_visible(False)
+plt.ylabel('AUPR difference')
+plt.title('Performance difference between old and new filtering scheme', size=14)
+
+
+# The plots above show the difference for each gene between the "old" filtering scheme (per-cancer type) and the "new" filtering scheme (gene-level across all cancer types). A positive value for a gene means that gene's classifier performed better for the "old" filtering scheme and vice-versa for the "new" filtering scheme.
+# 
+# We can see that most genes have a positive old vs. new difference, indicating that filtering for each cancer type independently (the "old" approach) tends to lead to better classifier performance for most genes. This seems to hold across all the data types we looked at.
```
arielah, line 300: It's a little weird to say "We can see that most genes have a positive old vs. new difference" right after a bar plot with equal numbers of classifiers in the positive and negative directions. I'd make it more obvious that the bar plot is looking at the tails, and if you want to you can also make a note that the most positive tail is bigger than the most negative tail.

02_classify_mutations/run_mutation_classification.py
```diff
@@ -212,6 +216,7 @@ def process_args():
                     gene,
                     classification,
                     gene_dir,
+                    filter_cancer_types=(not model_options.filter_all_data),
```
arielah, line 219: This notation is weird to me. Why not just have a parameter called --filter_by_cancer_types?

mpmp/utilities/tcga_utilities.py
```diff
@@ -40,7 +41,8 @@ def process_y_matrix(y_mutation,
     filter_count: the number of positives or negatives required per cancer-type
     filter_prop: the proportion of positives or negatives required per cancer-type
     output_directory: the name of the directory to store the gene summary
-    filter_cancer_types: if False, don't filter cancer types that don't meet criteria
+    apply_filter: if False, don't filter samples/cancer types at all
```
arielah, line 44: Would you ever want to do this?

arielah (approved):
Looks good! The reviewer was probably right that it's good to check how it works this way, and now you know. I just had a couple of comments to help the future readers of this code once this is published.

02_classify_mutations/nbconverted/compare_filtering.py
jjc2718: Good point! That line was mostly referring to the box plot (where the means are all greater than 0) but you're right that it would be clearer if I elaborate a bit.

02_classify_mutations/run_mutation_classification.py
jjc2718: I wanted the default to be what we were doing before (and will likely continue to do after this), so I didn't want to add a parameter to these scripts that I'd then have to go back and add to all my bash/cluster scripts, etc. Having a flag to turn on the "new" filtering version made more sense at the script level.

I guess I could have the function also take the `filter_all_data` argument, which would make this logical not/negation unnecessary. Maybe that would make more sense.

02_classify_mutations/nbconverted/compare_filtering.py
jjc2718: Yeah, it is a bit surprising. Maybe some of it is just threshold choice since a lot of points seem to be clustered below the 0.001 dotted line, but I don't think that necessarily explains it. I'll have to think about it.

PR #86, opened 2022-03-28
Predictions on samples from normal tissue

jjc2718:
One of our reviewers asked about our model's predictions on the TCGA samples from healthy tissue, which are present for some cancer types but not others (see [`01_explore_data/normal_tissue_samples.ipynb`](https://github.com/jjc2718/mpmp/blob/cb961db7714d68463cfb33b85d4641f39416771f/01_explore_data/normal_tissue_samples.ipynb) for some exploration of this).

This PR adds code to train and save a model on the whole (tumor) dataset for a given gene, and a notebook ([`07_train_final_classifiers/predict_normal.ipynb`](https://github.com/jjc2718/mpmp/blob/cb961db7714d68463cfb33b85d4641f39416771f/07_train_final_classifiers/predict_normal.ipynb)) to apply these trained models to the normal samples and compare their predictions to the predictions made on tumor samples.

For now, we just did this for the genes we looked at in the multi-omics analysis: TP53, KRAS, EGFR, SETD2, PIK3CA, IDH1. In most cases we do see that normal samples are predicted to have a low-ish probability of mutation. Here are the results for TP53 as an example:

image

So predictions for the normal samples are generally lower than they are for the mutated samples (true positives), but not as low as they are for the non-mutated tumor samples (true negatives). This makes some sense since the normal samples are "out of distribution" (they weren't present in the training set and probably don't exactly "look like" any of the tumor samples), so it makes sense that they have mutation probabilities closer to 0.5 than the "true" negative tumor samples that we fit the model on.

**Code changes:**

* Analysis notebooks described/linked above
* `07_train_final_classifiers/train_classifier.py`: script to train a classifier for a single gene/set of parameters on the _entire_ dataset (no cross-validation), and to save the model to `results_dir`
* `mpmp/utilities/param_results_utilities.py`: set of functions to get the "best" parameter choices from a set of results across different outer cross-validation folds
* Changes to `mpmp/prediction/cross_validation.py` and `mpmp/prediction/classification.py` to train a single model on the whole dataset


07_train_final_classifiers/train_classifier.py
```diff
@@ -0,0 +1,230 @@
...
+              file=sys.stderr)
+        log_df = fu.generate_log_df(
+            log_columns,
+            [gene, model_options.training_data, 'gene_not_found']
+        )
+        fu.write_log_file(log_df, io_args.log_file)
+        exit()
+
+    # train model
+    # TODO: save parameter selection
```
ben-heil, line 187: Does this still apply?

mpmp/prediction/classification.py
```diff
@@ -395,14 +416,16 @@ def get_preds(X_test_df, y_test_df, cv_pipeline, fold_no):
...
 
-    # make sure we're actually looking at positive class prob
-    assert np.array_equal(cv_pipeline.best_estimator_.classes_,
-                          np.array([0, 1]))
+    try:
+        # make sure we're actually looking at positive class prob
+        assert np.array_equal(cv_pipeline.best_estimator_.classes_,
+                              np.array([0, 1]))
+    except AttributeError:
+        assert np.array_equal(cv_pipeline.classes_, np.array([0, 1]))
```
ben-heil, line 428: I'm not sure when the `best_estimator_` attribute is expected to exist, but it may be clearer to check for it explicitly rather than by catching an AttributeError

ben-heil (approved):
Looks good! Makes sense that normal samples would kind of fall randomly along the axis, though I'd expect the sigmoid/the model being linear to push random data towards one extreme or the other. 

I'm not sure of the context of the reviewer comment, but if your goal is to say "the normal samples are out of distribution", then running a quick UMAP or PCA to show that the normal samples live in a different world may support your case

mpmp/prediction/classification.py
jjc2718: Makes sense - I was trying to go for [Python-style EAFP rather than LBYL](https://docs.python.org/3.8/glossary.html#term-eafp), but I agree that a comment would probably make things a bit clearer here.

jjc2718:
> I'm not sure of the context of the reviewer comment, but if your goal is to say "the normal samples are out of distribution", then running a quick UMAP or PCA to show that the normal samples live in a different world may support your case

This is a good idea! I'll do this as part of my next PR.

PR #88, opened 2022-05-02
Multi-omics neural network models, part 2

jjc2718:
This PR plots the results/comparison between the three-layer NN and elastic net for the multi-omics mutation prediction problem. Original implementation was done in #84, so there's no huge changes to the core modeling code here, just data processing and visualization.

We do see an improvement for multi-omics with PIK3CA, but not for the other genes:

multi_omics_best_model

And in general, NN performance is slightly better than the elastic net in a few cases (TP53, EGFR sort-of), but mostly the same or worse:

multi_omics_compare_mlp


mpmp/utilities/plot_utilities.py
```diff
@@ -602,16 +602,34 @@ def plot_multi_omics_results(results_df,
...
+
+        if model_compare:
+            if ix != 0:
+                ax.get_legend().remove()
+            if ix >= 3:
+                ax.set_xlabel('Training data type', size=13)
+            else:
+                ax.set_xlabel('')
+            for tick in ax.get_xticklabels():
+                tick.set_rotation(60)
```
ben-heil, line 624: I'm not sure what this does, can you add a comment?

ben-heil (approved):
Looks good!

05_classify_mutations_multimodal/compare_mlp.ipynb
```diff
@@ -1,5 +1,14 @@
 {
```
ben-heil, line 1: <p>What is the difference between <code>mlp, random</code> and <code>mlp, random, 5000 PCs</code>? It seems like only using PCs seems to really tank the MLP performance, but maybe that setting makes it more comparable to the elastic nets?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/mpmp/pull/88/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

ben-heil, line 1: <p>It never ceases to amaze me how much bigger model variance is than the effect of different model types </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/mpmp/pull/88/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

jjc2718: <p><code>mlp, random</code> in the old results was with 1000 PCs. We ended up not using those results because like you said it's harder to make the case that it's exactly comparable to the elastic net setup (those models use 5000 PCs), although the performance is generally better, presumably because the 5000 PC NNs end up overfitting somewhat.</p><p> </p><p>It didn't change the comparative results (in <code>05_classify_mutations_multimodal/plot_multimodal_results.ipynb</code>)to use 1000 PCs, so we decided not to explore it any further, even though the individual performance for 1000 PCs was more comparable to the elastic net models.</p>

jjc2718: <p>Right! We've seen this a lot with the mutation prediction problem (and probably lots of other problems are this way too, especially in biology).</p>

PR #91, opened 2022-05-05
Small (hopefully last) figure tweaks

jjc2718:
This should be the last of my many branches that I used to make tweaks to the figures for our paper revision. Changes are pretty minimal, they include fixing the heatmap for methylation data (one of our supplementary figures) and updating the code to plot raw vs. compressed results for a few genes (Figure 4 in the paper).


02_classify_mutations/plot_methylation_results.ipynb
```diff
@@ -876,7 +876,7 @@
     },
```
ben-heil, line 876: <p>Why is there a conditional statement around the plot rather than having two different cells or something? I feel like I've asked this before so sorry for asking it again.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/mpmp/pull/91/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

ben-heil (approved):

02_classify_mutations/plot_methylation_results.ipynb
jjc2718: <p>All of the results in the notebook will be different if <code>merged_geneset</code> is True, not just this one. The only difference here is that the plot is an unreadable mess with all the added genes, so the conditional plots it as two rows which is much more readable.</p><p> </p><p>Along those lines I guess I could have two sections of the notebook, one that plots the old Vogelstein gene set results and one that plots the new results in the paper, but my logic is that having the variable at the top that defaults to True will be less confusing to new users (i.e. people who just want to reproduce the results of the paper and don't know or care about the fact that at one point we used a different gene set that gave us different results).</p>

ben-heil: That make sense, thanks!

PR #92, opened 2022-07-01
Larger figure fonts for publication

jjc2718:
This looks like a decent number of changes, but all I'm doing here is making the font size bigger for all the figures, which the journal wanted for the published version of the paper.

To make the font size bigger in the heatmaps I had to split them into 3 rows instead of 2, which accounts for most of the changes to `mpmp/utilities/plot_utilities.py`.


arielah (approved):
Looks good! Congrats on it getting published soon!

PR #89, opened 2022-05-02
Save coefficients/hyperparameters for gene expression models

jjc2718:
As part of our revised manuscript (greenelab/mpmp-manuscript#46), we wanted to provide the coefficients (elastic net effect sizes) and hyperparameters (elastic net regularization strength + L1/L2 ratio) for the fit models using gene expression, for each gene in our cancer gene set; this PR fits the models on all the data and saves the coefficients/params.

Main additions in this PR:

* Bash script to fit models for all genes: `07_train_final_classifiers/scripts/run_all_genes.sh`
* Notebook to assemble model info into `.tsv` files: `07_train_final_classifiers/collect_full_models.ipynb`
* Coefs/params files saved in `data/final_models`

PR #93, opened 2022-07-01
Get final models (coefficients) for all data types

jjc2718:
As part of the published version of the paper, we made available trained models for all the data types we looked at, for all of the genes from our "merged" cancer gene set. (Before, after the first revision, we were just doing this for gene expression - see PR #89).

This PR just generalizes `07_train_final_classifiers/collect_full_models.ipynb` to save models (coefficients and parameters) for all data types rather than just expression, and extends `07_train_final_classifiers/scripts/run_all_genes.sh` to also run for all data types.


jjc2718:
No rush on this, I know you're out this week/will probably be busy when you return

ben-heil (approved):
Looks good! Sorry for the delay

